PR #23, opened 2019-08-08
So3g jupyter docker

jlashner:
This PR adds the `so3g-jupyter` docker to this repository, and makes it so that its automatically built with the most up to date version of so3g on each commit. It also has ocs installed, so it can be used to interactively start and stop agent operations.


.travis.yml
```diff
@@ -26,16 +26,17 @@ jobs:
...
+        - docker-compose build
+
         # Test import within container
         - docker run --rm so3g /usr/bin/python3 -c 'import so3g'
 
       after_success:
         # Tag all images for upload to the registry
-        - "docker tag so3g:latest ${DOCKERHUB_ORG}/so3g:latest"
-        - "docker tag so3g:latest ${DOCKERHUB_ORG}/so3g:${DOCKER_TAG}"
+        - "docker-compose config | grep 'image: ' | awk -F ': ' '{ print $2 }' | xargs -I {} docker tag {}:latest ${REGISTRY_URL}/{}:latest"
```
BrianJKoopman: `REGISTRY_URL` should still be `DOCKERHUB_ORG`. These are environment variables set on the Travis website. `REGISTRY_URL` is the old one I was using.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.2'
+services:
+
+  so3g:
+    image: "so3g"
+    build: .
+
+  so3g-jupyter:
+    image: "so3g-jupyter"
+    build: ./so3g-jupyter/
```
BrianJKoopman: I think we should put this, and other potential Dockerfile containing directories into their own directory, perhaps `docker/`, to keep the top level of the repo from getting cluttered.

PR #25, opened 2019-08-16
Cast data and timelines to native python types

BrianJKoopman:
Data within fields and timelines returned from `get_data()` currently return
`spt3g.core.DoubleVector`. To make it easy to work with loaded data cast these
as lists when we load the data.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -210,12 +210,13 @@ class HKArchive:
         timelines = {}
         for group_name, block in blocks_out:
             timelines[group_name] = {
-                't': block.t,
+                't': np.array(block.t),
                 'finalized_until': block.t[-1],
                 'fields': list(block.data.keys()),
             }
             for k,v in block.data.items():
-                data[k] = v
+                data[k] = np.array(v)
+
         return (data, timelines)
 
     def simple(self, fields=None, start=None, end=None, min_stride=None,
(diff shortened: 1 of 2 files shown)


mhasself:
Please emit numpy arrays instead of lists.  "native" python types are not well-suited to numerical applications.

BrianJKoopman:
Right. I was motivated by the component I was passing the resulting dictionaries into in sisock not liking the DoubleVector or np.array types, seems I won't be able to avoid casting to a list before doing so then. I'll switch to np.array's, test, and push a new commit.

mhasself:
@BrianJKoopman Is this still an issue?

BrianJKoopman:
No, since I had to cast the array in question to a list anyway (this was for proper encoding for when data was getting passed around by sisock) I did it in the code I was working on. Sorry for not following up with the np.array change, would we still prefer that over `spt3g.core.DoubleVector`?

mhasself:
Ahh... sorry I was reading the code wrong.  Yes, please, np.array.  This is basically a bug-fix; the docstring already claims it will return an array.

BrianJKoopman:
Switched to numpy arrays. Wrote a simple test case for verifying the type too. While simple, I hope it'll form the basis for other tests for `hk.getdata.py`.

Issue #24, opened 2019-08-15
hkcs.process_file() needs to be passed files in chronological order


BrianJKoopman:
I'm finding some strange behavior when I build a list of files to look at with the HKArchiveScanner dynamically and pass that list to hkcs.process_file one by one.

I'm building the list with an os.walk:
```
search = os.walk(data_dir)
files = []

for root, directory, _file in search:
    for f in _file:
        if f[-2:] == 'g3':
            files.append(os.path.join(root,f))
```

This produces an unsorted list, for example:
```
['/app/data/15657/sub/2019-08-14-00-29-31.g3', 
 '/app/data/15657/sub/2019-08-13-22-29-27.g3', 
 '/app/data/15657/sub/2019-08-14-01-29-32.g3', 
 '/app/data/15657/sub/2019-08-14-04-29-38.g3', 
 '/app/data/15657/sub/2019-08-13-21-29-25.g3', 
 '/app/data/15657/sub/2019-08-14-02-29-34.g3', 
 '/app/data/15657/sub/2019-08-13-20-29-24.g3', 
 '/app/data/15657/sub/2019-08-13-19-29-22.g3', 
 '/app/data/15657/sub/2019-08-13-18-29-20.g3', 
 '/app/data/15657/sub/2019-08-14-03-29-36.g3', 
 '/app/data/15657/sub/2019-08-14-05-29-39.g3', 
 '/app/data/15657/sub/2019-08-13-23-29-29.g3']
```

When passing to hkcs.process_file one by one and plotting the result much of the data is dropped:
![image](https://user-images.githubusercontent.com/7691438/63126244-7407db80-bf7d-11e9-89be-59c40d1f231f.png)

If one sorts the list first though, then all of the data is present in the resulting HKArchiveScanner:
![image](https://user-images.githubusercontent.com/7691438/63126280-8aae3280-bf7d-11e9-84f5-6790987125e4.png)

If this is a requirement for properly loading the data, we should probably document it somewhere.

**so3g version:** 0.0.4+43.g037ed0c

PR #26, opened 2019-08-16
Fix channel scramble

mhasself:
This is intended to address #24 , @BrianJKoopman please check.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -399,6 +399,10 @@ class HKArchiveScanner:
         optional index_info argument set to a dictionary containing
         the filename and byte_offset of the frame.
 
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Passing in an unsorted list still results in the plot missing data.

image

```
>>> so3g.__version__
'0.0.4+45.gef6fe65'
```

Sorting does solve the problem still.

image

mhasself:
Ok... tested on those data now.  I think this fixes it (https://github.com/simonsobs/so3g/pull/26/commits/5e966acaf641a7e1f2ba1cf4aa660254172e689c).

BrianJKoopman:
Yup, that did it -- passing an unsorted list of files now works.

PR #31, opened 2019-09-18
created g3 reader

laurensaunders:
New g3 reader in python package. Current functionality for reading SMuRF g3 data to an array. (Same as smurf_g3_reader from socs)

### DIFF
--- a/python/reader.py
+++ b/python/reader.py
@@ -1,0 +1,68 @@
+import so3g
+from spt3g import core
+import numpy as np
+import pickle
+import datetime, time
+import sys, os
+
+def g3_to_array(g3file):
+    """
+    Takes a G3 file output from the SMuRF archiver and reads to a numpy array.
+
+    Parameters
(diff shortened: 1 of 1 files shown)


nwhitehorn:
Do you want to structure this in terms of files? If you do it in terms of streams, you could do the processing on live data. G3File is really intended only for use in interactive sessions.

mhasself:
> Do you want to structure this in terms of files? If you do it in terms of streams, you could do the processing on live data. G3File is really intended only for use in interactive sessions.

Yes; this interface is targeting interactive / off-line data analysis.

nwhitehorn:
"Offline" is not quite the same as "interactive" -- there are some other benefits you get from the streaming mode (handling multiple files, parallel IO, etc.). But the requirement drops a bit.

BrianJKoopman:
Works now on recent files, merging!

Issue #27, opened 2019-08-16
New G3 Vector Types for SO?


mhasself:
This issue is intended to start a conversation about new data types we might want to have to store the SO TOD (both detector and ancillary streams) properly.  If some consensus is reached, the resulting work can be broken up and re-issued.

High-level concerns:

1.  FLAC provides very good compression for many of our signals, but the current G3Timestream forces the signal to lie in the dynamic range corresponding to 24-bit integers.  This is inconvenient in cases where the signal has a natural calibration that takes the meaningful part of the signal out of this range.  We thus desire to have FLAC compression and gain+offset transformations encoded in a vector object.

2.  G3Timestreams unpack to 64-bit floats.  This level of precision is basically never needed for anything.  It is definitely not needed to process detector data, which is the bulk of our data set.  Even if we just transform, on load, to float32, the additional RAM and processing overhead is a little silly.  So we want better support for unpacking timestreams directly to float32.

3.  Our Housekeeping data structure, G3IrregDouble, is ready for an upgrade.  The idea of storing timestamps co-sampled with 1 or more data vectors is good... but we need more flexibility in those data types.  Right now everything is uncompressed doubles, which is wasteful and actually has consequences because we sometimes want to look at HK on long times scales.  We should try to support, on equal footing, time-stamped streams of:
- float64, float32 (including with FLAC compression and gain+offset transforms).
- [u]int64/32/16/8.
- A string enumerator data type; for example to efficiently encode streams drawn from the set ["OK", "ERROR1", "ERROR2"].
- Strings.  The application here is log messages.

4.  Re-sampling algorithms that work on "normal" signals (detector timestreams, e.g.) do not work properly on signals that live on a periodic domain (such as HWP angles) or integer signals representing bit-masks (such as telescope status bits).  We should consider some form of annotation that would indicate, to a downstream consumer, that a field is periodic from [A, B).  For bit-masks, the equivalent minimal metadata would tell us what to do when trying to "down-sample" the sequence [0,1,1,0,0] into a single bit -- either we should AND those values together (to get 0 in this case), or OR them (to get 1).  (This isn't perfectly general but it would help us do the right thing much of the time.)

nwhitehorn:
A few quick comments on the first two items.

[...]

PR #32, opened 2019-10-01
G3VectorFixed: FLAC-ble G3VectorDouble with precision.

mhasself:
Like G3Timestream, G3VectorFixed has FLAC-based compression built in
to its serialization.  Unlike G3Timestream, the meaning of a single
unit in the compressed stream is now set by the "precision".  This
makes it easy to store values from a device that records data in fixed
units (e.g., 0.01 K).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,7 @@ add_library(so3g SHARED
   src/G3WCS.cxx
   src/G3Ndmap.cxx
   src/G3IndexedReader.cxx
+  src/G3VectorFixed.cxx
   src/Intervals.cxx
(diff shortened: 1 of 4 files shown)


mhasself:
This provides more flexibility in compressing vectors, as outlined in #27.  It's still a bit experimental... but I would like to use something like this for Housekeeping, soon, because it has compression.  

To the user this is still presented as a vector of doubles... I'm reluctant to collapse this to single-precision floats because the difference in the encodings means you see sizable rounding errors once you start to make use of the full dynamic range.

mhasself:
This is superseded by G3SuperTimestream, #98 .

PR #37, opened 2019-10-29
Some housekeeping for so3g.hk

mhasself:
@jlashner For OCS use, note the session_id can now be computed by HKSessionHelper.

### DIFF
--- a/test/test_hkagg.py
+++ b/test/test_hkagg.py
@@ -10,8 +10,8 @@ print('Streaming to %s' % test_file)
 w = core.G3Writer(test_file)
 
 # Create something to help us track the aggregator session.
-hksess = so3g.hkagg.HKSession(session_id=1234,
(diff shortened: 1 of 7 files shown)


test/test_hkagg.py
```diff
@@ -1,55 +1,87 @@
...
+        scanner = so3g.hk.HKScanner()
+        pipe = core.G3Pipeline()
+        pipe.Add(core.G3Reader(test_file))
+        pipe.Add(scanner)
+        pipe.Run()
+
+        print('Stats: ', scanner.stats)
+        print('Providers: ', scanner.providers)
+
+        assert (scanner.stats['concerns']['n_error'] == 0)
```
BrianJKoopman: Great to see tests! Since we're using unittest we should also use unittests's assertion methods: https://docs.python.org/3/library/unittest.html#assert-methods

PR #39, opened 2019-11-21
Handle spt3g quat being in core or coordinateutils

mhasself:


mhasself:
Self merging but please comment if you don't like the approach!

PR #43, opened 2020-02-20
add function to automatically load hk fields between to time ranges

kmharrington:
Adds function that just takes 2 datetime objects and will return the HK data between those objects. 

Uses a bunch of conventions I'm used to using.
1) Automatically looks for data in `$OCS_DATA_DIR` or can be given a data directory. I just made that up, is there another default way to reference that directory? 

2) Also uses `logging` for debugging messages, which I like but I don't know if that's something to add to so3g. Thoughts or alternatives?

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -452,7 +452,9 @@ def load_range(start, stop, fields=None, alias=None, data_dir=None):
     '''
     Args:
         start - datetime object to start looking
+                (should set tzinfo=dt.timezone.utc if your computer is not in utc)
         stop - datetime object to stop looking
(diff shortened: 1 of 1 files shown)


JakeSpisak:
FYI, I am currently using this code and can confirm it works

kmharrington:
I'm going to pause on merging until I've updated our system with the new DAQ file naming scheme (which will happen when we warm up early next week). I'll update this function to use that scheme and then test/merge.

kmharrington:
@JakeSpisak  would you mind testing the updated version (to deal with the new file names) on your setup? Thanks!

kmharrington:
Thanks Matthew, I'll merge this request and add the expansions as options.

PR #44, opened 2020-02-20
Reader tes bias and verbose

laurensaunders:


python/det/__init__.py
```diff
@@ -0,0 +1,3 @@
+import so3g
+from spt3g import core
+import numpy as np
```
mhasself: These imports shouldn't be here... the stuff in ```__init__``` isn't available in the other .py files, alas.  Rather ```__init__.py``` is used to gather certain things from local .py files into the main level of the submodule.  In our case we'd want

```
from .det_reader import g3_to_array
```

PR #46, opened 2020-03-03
Support readthedocs, including boost-python classes

mhasself:
docs/ now has code that scans a compiled version of so3g and produces
a pure-Python "shell" library with the same docstrings.  This code can
be substituted in for Sphinx to process on readthedocs.  Also some
helper functions and cmake target to cause these things to happen.


mhasself:
I should note that this works... see recent builds at https://so3g.readthedocs.io/en/latest/ .

Second... the details and the process by which one should push to readthedocs has not been fully documented anywhere.  This is rather tightly integrated with github so it makes sense to me to store such information on the github wiki for this project.  But I could be talked into something else, probably (in the Sphinx docs/? in the README?).

Issue #47, opened 2020-03-09
Expand load_range to take ctime and update docs


kmharrington:
From Matthews suggests on the first PR: 

> I think datetimes are nice, but it would be cool to have this accept, optionally, a couple of other formats for the arguments. A nice place to start might be this function:
def to_timestamp(arg):
    """Convert arg to a unix timestamp as float.  Accepts datetime, float,
    or int arguments.  This could be expanded to convert strings as well!
    """
    if isinstance(arg, dt.datetime):
        return arg.astimezone(dt.timezone.utc).timestamp()
    return float(arg)
A system for publishing so3g docs is now in development... can you add a reference to this function in docs/hk.rst, so that the docstring will (eventually) be available at https://so3g.readthedocs.io/en/latest/hk.html ?

PR #48, opened 2020-03-12
Expand load range functionality

kmharrington:
Solves issue #47 . 

`load_range` will work with several different types of input date format and I added a configuration file option because that's how I want to use it.

Also wrote a lot of docs.

### DIFF
--- a/docs/hk.rst
+++ b/docs/hk.rst
@@ -11,6 +11,126 @@ If you're here you probably just want to read in some data, so we will
 start with that.  Later on we go into details of the data model and
 the interfaces for writing compliant HK files.
 
+Loading Data Saved OCS/SO Style
+---------------
+
+One of the most basic things we might want to do is load data between
+a time range. For `.g3` files that are saved by an OCS Aggregator, there
+is a specific folder structure and file naming scheme. ``so3g.hk.load_range``
+is writen to load data for a specified time frame saved in that style.
+
+Example Use::
+
+    from so3g.hk import load_range    
+    data = load_range(start, stop, **kwargs)
+
+Defining Time Ranges
+```````````````````````
+
+There are several options for defining start and stop. These options are passed
+to ``so3g.hk.to_timestamp(some_time)`` to be parsed.
+
+* ``datetime.datetime`` objects::
+
+    import datetime as dt
+    # tzinfo is likely needed if your computer is not in UTC
+    start = dt.datetime(2020, 3, 12, 1, 23, tzinfo=dt.timezone.utc)
+    
+* Integers or Floats - these are assumed to be ctimes 
+
+* Strings - assumed to be UTC dates and parsed by ``datetime.strptime``
+
+    * '%Y-%m-%d'
+    * '%Y-%m-%d %H:%M'
+    * '%Y-%m-%d %H:%M:%S'
+    * '%Y-%m-%d %H:%M:%S.%f'
+    * Or submit your own with the ``str_format`` argument
+
+Define Where to Look for Data
+``````````````````````````````````````````````
+
+* **Option 1** 
+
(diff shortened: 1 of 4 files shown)


kmharrington:
Looks like the build is failing because of `import pytz`, which I added to fix some timezone issues. I think that's a pretty standard package? It was installed by default with either python or anaconda3.

mhasself:
I believe you need to add that package to the "requirements.txt" file.  Packages listed there are installed through pip prior to building + testing.  Since it's pip, use the pip name... which appears to be pytz.

BrianJKoopman:
You're also missing the `pyaml` package.

kmharrington:
We'll get this eventually! 🥇

python/hk/getdata.py
```diff
@@ -448,7 +451,38 @@ def __init__(self, prefix, fields, start, end, index_info):
         self.index_info = index_info
 
         
-def load_range(start, stop, fields=None, alias=None, data_dir=None):
+def to_timestamp( some_time, str_format=None ):
```
mhasself: No spaces around args, please... to_timestamp(some_time, ...).

docs/hk.rst
```diff
@@ -11,6 +11,126 @@ If you're here you probably just want to read in some data, so we will
 start with that.  Later on we go into details of the data model and
 the interfaces for writing compliant HK files.
 
+Loading Data Saved OCS/SO Style
+---------------
```
mhasself: Banner underline too short... sphinx actually complains about this.  If in emacs, try rst-mode magic:  C-c C-a C-a.

docs/hk.rst
```diff
@@ -11,6 +11,126 @@ If you're here you probably just want to read in some data, so we will
...
+is a specific folder structure and file naming scheme. ``so3g.hk.load_range``
+is writen to load data for a specified time frame saved in that style.
+
+Example Use::
+
+    from so3g.hk import load_range    
+    data = load_range(start, stop, **kwargs)
+
+Defining Time Ranges
+```````````````````````
```
mhasself: Line too long (and more below)!  Sphinx doesn't complain about this... but I do.  C-c C-a C-a ?

python/hk/getdata.py
```diff
@@ -448,7 +451,38 @@ def __init__(self, prefix, fields, start, end, index_info):
...
+    Args:
+        some_time - if datetime, converted to UTC and used
+                    if int or float - assumed to be ctime, no change
+                    if string - trys to parse into datetime object
+                              - assumed to be in UTC
+        str_format - allows user to define a string format if they don't
+                    want to use a default option
+    Returns:
+        ctime of some_time
+    '''
```
mhasself: This docstring (and the load_ranges one) isn't quite numpy style.  That's ok for now, we can fix it up later (when building + posting of docs is more established).

python/hk/getdata.py
```diff
@@ -448,7 +451,38 @@ def __init__(self, prefix, fields, start, end, index_info):
...
+        str_options = ['%Y-%m-%d', '%Y-%m-%d %H:%M', '%Y-%m-%d %H:%M:%S', '%Y-%m-%d %H:%M:%S.%f']
+        for option in str_options:
+            try:
+                return to_timestamp(pytz.utc.localize( dt.datetime.strptime(some_time, option )))
+            except:
+                continue
+        hk_logger.error('Could not process string into date object, options are: {}'.format(str_options))
+        return
+    hk_logger.error('Type of date / time indication is invalid, accepts datetime, int, float, and string')
+
```
mhasself: Perhaps raise ValueError on parsing failure?  I think that's what I'd expect to happen.

kmharrington:
I might have made some git mistakes.... 

But I think I did fix all the comments 🥇

PR #49, opened 2020-04-03
Adding xieta functions in the quat.py file.

Zhilei:
Added the rotation_xieta() and decompose_xieta(), also switched the variable names between psi and gamma

### DIFF
--- a/python/proj/quat.py
+++ b/python/proj/quat.py
@@ -18,15 +18,19 @@ DEG = np.pi/180
 def euler(axis, angle):
     """
     The quaternion representing of an Euler rotation.
+    
     For example, if axis=2 the computed quaternion(s) will have
     components:
-      q = (cos(angle/2), 0, 0, sin(angle/2))
+
+        q = (cos(angle/2), 0, 0, sin(angle/2))
+    
     Parameters
     ----------
     axis : {0, 1, 2}
         The index of the cartesian axis of the rotation (x, y, z).
     angle : float or 1-d float array
         Angle of rotation, in radians.
+    
     Returns
     -------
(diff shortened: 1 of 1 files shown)


python/proj/quat.py
```diff
@@ -47,45 +43,49 @@ def euler(axis, angle):
     return G3VectorQuat(q)
 
 
-def rotation_iso(theta, phi, gamma=None):
+def rotation_iso(theta, phi, psi=None):
     """Returns the quaternion that composes the Euler rotations:
-
-        Qz(phi) Qy(theta) Qz(gamma)
-
+        Qz(phi) Qy(theta) Qz(psi)
```
mhasself, line 48: Please restore the blank lines, here and elsewhere.  Our docstrings are in numpy format, and ReST requires blank lines around indented blocks.

python/proj/quat.py
```diff
@@ -47,45 +43,49 @@ def euler(axis, angle):
...
-    return rotation_iso(np.pi/2 - lat, lon, gamma)
-
+    return rotation_iso(np.pi/2 - lat, lon, psi)
+
+def rotation_xieta(xi, eta, gamma=0):
+    """Returns the quaternion that rotates the center of focal 
+    plane to (xi, eta, psi)
+          xi = - sin(theta) * sin(phi)
+          eta = - sin(theta) * cos(phi)
+          psi = gamma - phi
```
mhasself, line 69: Again, need blank lines here.  And please intend these expressions to the same level as the other rotation_* docstrings.  Also you have left out the bit where about this rotation being equivalent to rotation_iso(theta, phi, psi) where ...

PR #54, opened 2020-04-17
More range updates

kmharrington:
More functionality for Ranges and RangesMatrix. 
1) Addition for RangesMatrix
2) `buffer` and `buffered` for expanding Ranges by set amounts
3) `close_gaps` to combine any ranges within a set amount

### DIFF
--- a/src/Ranges.cxx
+++ b/src/Ranges.cxx
@@ -146,6 +146,36 @@ Ranges<T> Ranges<T>::buffered(const T buff)
     output.cleanup();
     return output;
 }
+
+//Close gaps between Ranges if they are leq gap
+template <typename T>
+Ranges<T>& Ranges<T>::close_gaps(const T gap)
+{
+    auto p = segments.begin();
+    while (p != segments.end()) {
+        // Check for distance from the front
+        if (p->first <= gap){
+            p->first = 0;
+        }
+        if (p->second >= count-gap){
+            p->second = count;
+        }
+        // Check for distances from the next interval.
+        auto q = p+1;
+        if (q == segments.end())
+            break;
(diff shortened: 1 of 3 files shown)


src/Ranges.cxx
```diff
@@ -120,6 +120,62 @@ Ranges<T>& Ranges<T>::merge(const Ranges<T> &src)
     return *this;
 }
 
+//Buffer Range in place
+template <typename T>
+Ranges<T>& Ranges<T>::buffer(const T buff)
+{
+    auto p0 = this->segments.begin();
+    while (p0 != this->segments.end()){
```
mhasself, line 128: Throughout... can you add the space between `){` to make it `) {`?

src/Ranges.cxx
```diff
@@ -672,6 +728,17 @@ using namespace boost::python;
          return_internal_reference<>(),                                 \
          args("self", "src"),                                           \
          "Merge ranges from another " #CLASSNAME " into this one.")     \
+    .def("buffer", &CLASSNAME::buffer,                                  \
+        return_internal_reference<>(),                                  \
+        args("self", "buff"),                                           \
+        "Buffer each interval by an amount specificed by buff")         \
```
mhasself, line 753: speeling

kmharrington:
Done! Also I added ~2 more functions to `RangesMatrix`

mhasself:
Fun... can I advocate for renaming to zeros_like and ones_like?  I think that's more consistent with numpy terminology (where empty means unitialized).  Also to avoid another comment, you need a line between your two new functions in RangesMatrix!

kmharrington:
Makes sense, updated!

PR #55, opened 2020-04-17
docs: describe process used to push to readthedocs

mhasself:


mhasself:
Yes, you can't get doc changes into readthedocs until they've been merged into master.

Developers can still edit and build docs locally to confirm formatting looks ok; the mocking up of C++ class docstrings isn't done if you just do "cd docs; make html".

mhasself:
But perhaps you were asking if you had to already have master checked out to go through this process?  You sort of do, because master has to be installed.  Just added a note about this.

github requires you to re-approve, sry!

BrianJKoopman:
Ah, right, I see, you'd probably have worked on something, PR'd it, had it merged, then followed this on the master branch. Makes sense, thanks!

PR #57, opened 2020-05-12
Use new spt3g cmake helpers

mhasself:
This removes duplicate / conflicting code in so3g vs. spt3g CMakeLists, and ensures that so3g is built with the same stuff (e.g. boost) as spt3g.  I've tested this on two systems (including a shared installation) and it is quite smart.


mhasself:
@BrianJKoopman I assume the Travis passing means that this is not a problem for our docker builds... that might be a slight extrapolation.

BrianJKoopman:
> @BrianJKoopman I assume the Travis passing means that this is not a problem for our docker builds... that might be a slight extrapolation.

No, that's right. The images were built and tests run within the container.

PR #58, opened 2020-05-13
Support for HK format v1

mhasself:
Updates loaders, checkers, and testing to be compatible with current HK schema (v0) and a new schema (v1) that makes use of G3TimesampleMap.


jlashner:
Just checked this with a new data file from here: https://github.com/simonsobs/ocs/pull/128 and it works. Have you checked the translator with old data?

python/hk/translator.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self):
+        self.stats = {'n_hk': 0,
+                      'n_other': 0,
+                      'versions': {}}
+
+    def Process(self, f):
+        """Translates one frame to the target schema.  Irrelevant frames are
+        passed through unmodified.
+
```
BrianJKoopman, line 31: Missing "Args:" and "Returns:" sections in docstring.

BrianJKoopman: Part of what motivated this comment was seeing all the returns here return a list `[f]`, and where `Translator` gets called it typically has a `[0]` index immediately following. I'm probably missing something, but why returns lists all the time here?

python/hk/scanner.py
```diff
@@ -29,6 +31,7 @@ def __init__(self):
                 'n_error': 0,
                 'n_warning': 0
             },
+            'versions': {}, # Number of frames counted for each hkagg_version.
```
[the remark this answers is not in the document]
mhasself: Yes, for sure... fixing in a new commit.

python/hk/translator.py
mhasself: Docstring will be updated, good call.

The HKTranslator and HKScanner are both G3Pipeline-compatible, which means the Process/__call__ can return either a G3Frame object or a (possibly 0-len) list of frames.  But when using these things in non-G3Pipeline, it's better not to have those output types mixed.  The list is more general so I seem to have defaulted to it.  (I think I'm back-filling justification here... it's possible I was just confused about G3Modules for a while... but I think returning a list here is looking forward.)  Anyway, adding text.

python/hk/scanner.py
```diff
@@ -19,6 +19,9 @@ class HKScanner:
           and error (key ``n_error``) events encountered.  The detail
           for such events is logged to ``spt3g.core.log_warning`` /
           ``log_error``.
+        - ``versions`` (dict): The number of frames (value) (value)
```
BrianJKoopman, line 22: (value) (value)

PR #63, opened 2020-06-17
Omnibus improvements to so3g.proj

mhasself:
Important new features in the so3g.proj system are:

- Tiled Maps: support for tiled maps, where a large footprint may be sub-divided into ~equally sized submaps, with only some useful subset of those in RAM for projection operations.
- Precomputed Pointing: support for projection operations using precomputed pointing, which can save a lot of time if repeated projections are needed and sufficient RAM is available.
- Detector Weights: support for weighting detectors with a scalar factor, when binning maps and weight maps, to facilitate inverse-variance weighting in simple noise models.
- Use of linear interpolation to accelerate inverse trig operations (asin and atan2).
- Better support for creating custom OMP thread assignments.

The following interfaces have been changed:
- EarthlySite.get_named has been removed; use so3g.proj.SITES dict instead.
- Projectionist interface is somewhat different, especially OMP support options.
- Internal ProjEng API is quite different
- BufferWrapper C++ class is now templated for data type and lives in numpy_assist.h

In addition, the following changes / improvements have been made:
- Documentation in proj.rst has been updated.
- tests/test_proj_eng.py - performs very basic tests of projection routines.
- tests/test_proj_quat.py - promoted to a proper unittest.TestCase


mhasself:
If no one volunteers to review this, I will merge in ~24 hours.  (Volunteering will stop the clock -- no need to complete the review in 24 hours!)

PR #64, opened 2020-06-20
docs: add more classes to extract_docstrings

mhasself:
These are accessed on import in hk for v1 support.


mhasself:
Self-merging.  This is innocuous except for readthedocs compat.

PR #67, opened 2020-07-16
hk format v2

mhasself:
As discussed on Slack, this brings in a field ('block_names') to unambiguously assign a name to each block.  Update to tod2maps_docs coming soon.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -343,7 +343,7 @@ class HKArchiveScanner:
             return [f]
 
         vers = f.get('hkagg_version', 0)
-        assert(vers == 1)
+        assert(vers == 2)
 
         if f['hkagg_type'] == so3g.HKFrameType.session:
             session_id = f['session_id']
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -46,10 +46,12 @@ _SCHEMA_V1_BLOCK_TYPES = [
 
 
 class HKArchive:
+    """Container for information necessary to determine what data fields
+    are present in a data archive at what times.  This object has
+    methods that can determine what fields have data over a given time
+    range, and can group fields that share a timeline (i.e. are
+    co-sampled) over that range.
 
-    """Contains information necessary to determine what data fields are
-    present in a data archive at what times.  It also knows how to
-    group fields that share a commong timeline.
     """
     field_groups = None
     def __init__(self, field_groups=None):
@@ -90,6 +92,13 @@ class HKArchive:
             fields belonging to this group.  ``fgs`` is a list of
             _FieldGroup objects relevant to the fields in this
             group.
+
+        Notes:
+            Each entry in the returned list carries information for
+            set of fields that are co-sampled over the requested time
+            interval.  Each of the requested fields will thus appear
+            in at most one group.
+
         """
         span = so3g.IntervalsDouble()
         if start is None:
@@ -297,12 +306,15 @@ class _HKProvider:
 
 
 class HKArchiveScanner:
-    """Consume SO Housekeeping streams and create a record of what fields
+    """Consumes SO Housekeeping streams and creates a record of what fields
     cover what time ranges.  This can run as a G3Pipeline module, but
     will not be able to record stream indexing information in that
     case.  If it's populated through the process_file method, then
     index information (in the sense of filenames and byte offsets)
     will be stored.
+
+    After processing frames, calling .finalize() will return an
+    HKArchive that can be used to load data more efficiently.
 
     """
     def __init__(self):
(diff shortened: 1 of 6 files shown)


ahincks:
Looking good!

I haven't traced through the code exhaustively but want to double check: in `getdata.py` does the use of `block_index` in the `index_info` only apply per frame? I.e., there is no assumption that blocks will appear in a certain order with a frame right? For instance, if a frame has two blocks, `block_a` and `block_b`, is it the case that in some frames the order is `block_a`, `block_b` but in others it could be `block_b`, `block_a`?

jlashner:
I'm pretty sure since dicts are now ordered in python, blocks will maintain order until a provider is reset, which happens when a feed isn't published to in a certain amount of time. Then when it comes back the blocks will be re-ordered based on which was published first.

mhasself:
Yes block_index is only valid per frame.  The order can change.  You are not even guaranteed to get both block_a and block_b in the same frame.

ahincks:
> Yes block_index is only valid per frame. The order can change. You are not even guaranteed to get both block_a and block_b in the same frame.

Good, that's what I expected. Thanks for clarifying.

mhasself:
> This will be great. The only annoyance will be in the use case where someone gets data over a time period where we switched from v0/v1 (and the translator is used) to v2, since the block name will change. But I think this is unavoidable and probably in practice won't be a big impediment.

The block name is a very soft guarantee.  There's definitely no guarantee that a given named block will always contain the same fields.  All that's assured is that group membership will be constant for the duration of a prov_id.  When the device re-registers the feed, it could be completely different (or just have some fields added or dropped).

The current 2-pass system in HKArchiveScanner / HKArchive is able to handle that gracefully -- if you ask for data only on a limited time range [t0,t1), it will identify the subsets of the fields that are co-sampled on that range.  Even if a lot of weird stuff happend between t0 and t1.  Down-sampling could use the same system, if it assumed the output of HKArchiveScanner was already available (I believe @jlashner  is working on that).

mhasself:
On the other hand... when talking about re-sampling, and the desire for the resampling algorithm to produce the same results no matter how the data are framed:  I think it's quite reasonable to forbid reframing that crosses prov_id boundaries.  Whatever algorithm you're running, it's ok to stop and restart when prov_id changes.  However, that prevent down-sampling to strides smaller than the lifetime of the prov_id... or those would have to be treated differently.  Not sure what to do here.

ahincks:
> However, that prevent down-sampling to strides smaller than the lifetime of the prov_id... or those would have to be treated 
differently. Not sure what to do here.

Yeah, it's questions like these (along with (re)learning `so3g`/`spt3g_software`) that are making writing the resampling take longer to prepare than anticipated. My plan is to make it go across multiple `prov_id` if requested. I'm slowly figuring this stuff out … and then we can debate them in a PR.

python/hk/translator.py
```diff
@@ -51,37 +67,62 @@ def Process(self, f):
...
+
+        if orig_version <= 1:
+            # Add 'block_names'.  Since we don't want to start
+            # caching Block Stream information, just compute a good
+            # block name based on the alphabetically first field in
+            # the block.
+            block_names = []
+            for block in f['blocks']:
+                field_names = list(sorted(block.keys()))
+                block_names.append('block_for_%s' % field_names[0])
```
BrianJKoopman, line 121: What's the implication for translating old data that doesn't have any block names and working with this data and new data that does have some set block name? i.e. you have some old Lakeshore data that got translated and the block name is "block_for_Channel_01_T", while in newer files it's something else like "temperatures".

ahincks:
Another idea is occurring to me: does it make sense for the HKStatus frame to announce what the names of all the block streams will be for each provider? E.g., to have `f['providers']['block_streams']` be a list of all the block streams that (may) be present in this provider? This would be handy programmatically.

Or is it the case that providers don't necessarily know ahead of time what block streams they will have?

mhasself:
> Another idea is occurring to me: does it make sense for `HKStatus` frames to announce what the names of all the block streams will be for each provider? E.g., to have `f['providers']['block_streams']` be a list of all the block streams that (may) be present in this provider? This would be handy programmatically.
> 
> Or is it the case that providers don't necessarily know ahead of time what block streams they will have?

It is the case.  The rules currently permit a block stream to start and stop whenever, as long as field membership is fixed.

python/hk/getdata.py
```diff
@@ -373,24 +369,19 @@ def Process(self, f, index_info=None):
...
+                                          'index_info': []}
                 # To ensure that the last sample is actually included
                 # in the semi-open intervals we use to track frames,
                 # the "end" time has to be after the final sample.
-                prov.blocks[rep]['end'] = b.times[-1].time / core.G3Units.seconds + SPAN_BUFFER_SECONDS
-                prov.blocks[rep]['index_info'].append(index_info)
+                prov.blocks[bname]['end'] = b.times[-1].time / core.G3Units.seconds + SPAN_BUFFER_SECONDS
+                ii = {'block_index': bidx}
+                ii.update(index_info)
+                prov.blocks[bname]['index_info'].append(ii)
```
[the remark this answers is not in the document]
mhasself: Yes... I am clarifying this in docstrings + comments, will push some more docs in a moment.

python/hk/translator.py
mhasself: One should not use block name for anything, except to group fields within the lifetime of a single prov_id.  So the answer should be "the implication is that this won't affect anything".  For example, the block_name shouldn't be visible to the user; and similarly any code that is trying to figure out co-sampled groups of fields should not use it beyond the scope of a single prov_id.

PR #68, opened 2020-07-21
SmurfArchive system

jlashner:
This PR introduces the system for indexing and efficiently loading smurf data.

The SmurfArchive system takes an archive_path and a db_path, and writes all the frame and file indexing info to an sqlite database at the specified path. The load_data function uses the index info and the G3IndexedReader to efficiently load detector and bias data between the requested start and stop times.

The `load_status` function can be used to load the current status at a specific point in time, however there is a lot more work that can be done to improve this, such as ignoring and squashing status frames where nothing has actually changed. It might be useful to index which fields in each status frame have actually changed, so people can request specific status fields without reading in every status frame. Also I think we should add a `status_timeline` function which would return a timeline for each field and the timestamps which the field changed.

This also probably needs tests, but it's been working really well for simons1 analysis so far.

### DIFF
--- a/docs/smurf_archive.rst
+++ b/docs/smurf_archive.rst
@@ -2,7 +2,15 @@ Loading Detector Data
 ======================
 
 The SmurfArchive module contains code for indexing and quickly loading detector
-data from an archive.
+data from an archive of streamed data written by the smurf-recorder OCS agent.
+A `data archive` should contain all timestream data recorded on a given system.
+The file paths within a data archive will look like::
+
+    <archive_path>/<date-code>/<stream-id>/<filename>.g3
+
+For each stream-id, it is important that the time ranges of different files do
+not overlap or else data may not be returned in the correct order.
+
 
 Indexing Timestream Data
 --------------------------
@@ -13,7 +21,7 @@ is needed for each archive. The following snippet will write an index database
 to ``<archive_path>/frames.db`` or update an existing database with new frames
 if it already exists::
 
-    from so3g.smurf.SmurfArchive import SmurfArchive
+    from so3g.smurf import SmurfArchive
     arc = SmurfArchive('/data/timestreams')
     arc.index_archive()
 
@@ -29,14 +37,15 @@ Loading Data
 
 To load detector data that includes the range ``[start, stop]``, use::
 
-    from so3g.smurf.SmurfArchive import SmurfArchive
+    from so3g.smurf import SmurfArchive
     arc = SmurfArchive('/data/timestreams')
     times, data = arc.load_data(start, end)
 
 The following will load the a dicitonary of the ``status`` variables at a
 specified time::
 
-    from so3g.smurf.SmurfArchive import SmurfArchive
+    from so3g.smurf import SmurfArchive
+    arc = SmurfArchive('/data/timestreams')
     arc = SmurfArchive('/data/timestreams')
     status = arc.load_status(time)
 
--- a/docs/smurf_archive.rst
+++ b/docs/smurf_archive.rst
@@ -6,10 +6,14 @@ data from an archive of streamed data written by the smurf-recorder OCS agent.
 A `data archive` should contain all timestream data recorded on a given system.
 The file paths within a data archive will look like::
 
-    <archive_path>/<date-code>/<stream-id>/<filename>.g3
+    <archive-path>/<date-code>/<stream-id>/<filename>.g3
 
 For each stream-id, it is important that the time ranges of different files do
-not overlap or else data may not be returned in the correct order.
+not overlap or else data may not be returned in the correct order. This current
+version of the SmurfArchive system only works for a single stream-id, however
+future changes to the SmurfRecorder OCS agent and the SmurfArchive class will
+provide support for archives containing timestreams from multiple stream-id's.
+This change will be necessary for systems with multiple smurf cards.
 
 
 Indexing Timestream Data
--- a/docs/smurf_archive.rst
+++ b/docs/smurf_archive.rst
@@ -1,0 +1,50 @@
+Loading Detector Data
+======================
+
+The SmurfArchive module contains code for indexing and quickly loading detector
+data from an archive.
+
+Indexing Timestream Data
+--------------------------
+To index data, the SmurfArchive module will walk through an archive and record
+information about each frame to an sqlite database, including the frame time,
+frame type, the number of channels and samples, etc. Only one index database
+is needed for each archive. The following snippet will write an index database
+to ``<archive_path>/frames.db`` or update an existing database with new frames
+if it already exists::
+
+    from so3g.smurf.SmurfArchive import SmurfArchive
+    arc = SmurfArchive('/data/timestreams')
+    arc.index_archive()
+
+In this example, all detector data stored in ``/data/timestreams`` will be
+indexed and the index database will be written to
+``/data/timestreams/frames.db``.
+
+A different database path can be specified by passing the ``db_path`` argument
+to the ``SmurfArchive``.
+
+Loading Data
+-------------
+
+To load detector data that includes the range ``[start, stop]``, use::
+
+    from so3g.smurf.SmurfArchive import SmurfArchive
+    arc = SmurfArchive('/data/timestreams')
+    times, data = arc.load_data(start, end)
+
+The following will load the a dicitonary of the ``status`` variables at a
+specified time::
+
+    from so3g.smurf.SmurfArchive import SmurfArchive
+    arc = SmurfArchive('/data/timestreams')
+    status = arc.load_status(time)
+
+See the API for descriptions of possible keyword arguments
+
+API
+---
+
+.. autoclass:: so3g.smurf.SmurfArchive.SmurfArchive
+   :members: index_archive, load_data, load_status
+
--- a/python/smurf/SmurfArchive.py
+++ b/python/smurf/SmurfArchive.py
@@ -13,11 +13,7 @@ import yaml
 
 
 num_bias_lines = 16
-type_key = {
-    core.G3FrameType.Observation: 0,
-    core.G3FrameType.Wiring: 1,
-    core.G3FrameType.Scan: 2
-}
+
 
 Base = declarative_base()
 Session = sessionmaker()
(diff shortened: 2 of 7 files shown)


ahincks:
I hope to take a look at this in the next couple of days!

jlashner:
Here's an example you can run on simons1:
```
from so3g.smurf.SmurfArchive import SmurfArchive
arc = SmurfArchive('/home/jlashner/data/ucsd-sat1/timestreams', db_path='/home/jlashner/frames.db')
arc.index_archive()
start = 1593046237.9455938 
end = start + 3*60*60
ts, ds, biases = arc.load_data(start, end, load_biases=True)
```

ahincks:
Your example code "worked out of the box" for me on `simons1`! Took 6 seconds to read in the 3 hours of data. Do you have a sense of what is dominating this time? (E.g., if this was housekeeping and I wanted 3 hours for just 10 thermometers, could I expect a significant speed-up?)

Here are a few suggestions from my first go-through:

#### Take full advantage of the relational database
- Instead of `frames.location` containing the filename and then having to do a split() on the colon, use two columns in the `frames` table: `file_id`, `idx`. The first is a foreign key to the ID of the file in the `file` table and the second contains the index number that is currently stored after the colon in the `location` column. You can easily (and efficiently) get the filename with a `JOIN`. (And note that SQLite does allow the `UNIQUE` parameter to be set for a pair of columns.)
- Make a `frame_type` table with a primary key `id` and `type` key that records `core.G3FrameType.Observation` etc. Then the `type` column in the `frames` table can be a foreign key referring to `frame_type`. This shouldn't affect performance and it will have the virtue that the DB is more self-describing.
  - (A general rule of thumb I use with relational DB's: if you have any kind of `type` column, make it refer to a table that defines what those types are—unless you can do with something like an enum (which SQLite doesn't support) and you're pretty sure it will always remain static.)

#### Think about how this index can be generalised to be used on HK data.
This may include more subtleties than I can see at the moment, but for starters:
- Add a table `file_type` that has an `id` and `type`; the latter for now would be `smurf` or `hk`. Then add a column to `files` that references this table.
- It would be great if the fields and Block Streams of HK data were recorded in this DB as well. Then you can query a time range to see if a given field is present.
  - This shouldn't be too hard if more tables are added; let me know if you want some specific ideas on how to structure the DB to do this.

ahincks:
P.S. `SQLAlchemy` and `tqdm` will need to be added to `requirements.txt`.

jlashner:
Hey Adam, thanks for the suggestions! Seems like some good ideas, I'll implement the relational database comments.

As for how this will behave for hk data, I think all of the time is spent reading each frame into memory, but with each frame being 1 sec, 3 hours worth of data is  ~10,000 frames. For HK, frames are more like 1 per 10 min, and are smaller in size, so 3 hours of hk data is more like 18 frames, and I believe each frame is smaller in size, so this should be much faster. I imagine the bottleneck in that case will be the sql query. 

I was kind of imagining that the smurf and hk indexes would be in different databases (one located in `<data_dir>/hk/index.db` and one in `<data_dir>/timestreams/index.db`. Even though the classes would be very similar, I'm not sure what the benefits would be of creating a general database and loader that includes both of them, especially since there is already an HKArchiveScanner class. 

As for what HK data should be indexed I think it might be a good idea to have a different row for each field segment, so we could do something like have three tables: 
1. `files` which just has the names of files that have been indexed
2. `frames` which would have the file id, frame idx within file, and offset
3. `segments` which would have: `field`, `provider`, `provider_session_id`, `frame_id` (from frames table), `start`, and `stop`.
But I'm not really sure if that discussion should be a part of this PR, we should continue talking on slack...

jlashner:
Finally got around to implementing Adam's suggestions and adding documentation currently hosted here: https://simons1.princeton.edu/~jlashner/so3g_docs/smurf_archive.html. Let me know if there's anything else I should add before merging! I'll wait until Adam gets back from vacation and has a chance to look it over.

python/smurf/SmurfArchive.py
```diff
@@ -0,0 +1,300 @@
...
+        session.close()
+
+    def load_data(self, start, end, show_pb=True, load_biases=False):
+        """
+        Loads smurf G3 data for a given time range. For the specified time range
+        this will return a chunk of data that includes that time range.
+
+        Args
+        -----
+        start (timestamp): start timestamp
```
mhasself, line 201: These args don't render properly in sphinx -- indent the list I think.

python/smurf/SmurfArchive.py
```diff
@@ -0,0 +1,300 @@
+import sqlalchemy as db
```
mhasself, line 1: Add sqlalchemy to the requirements file?  And sqlite, if not brought in by sqlalchemy...

docs/smurf_archive.rst
```diff
@@ -0,0 +1,50 @@
+Loading Detector Data
+======================
+
+The SmurfArchive module contains code for indexing and quickly loading detector
+data from an archive.
```
mhasself, line 5: What is "an archive"?  How do the files need to be organized / named?  What happens if the "archive" has multiple files that claim to have samples covering the same time ranges?

docs/smurf_archive.rst
```diff
@@ -0,0 +1,50 @@
...
+
+A different database path can be specified by passing the ``db_path`` argument
+to the ``SmurfArchive``.
+
+Loading Data
+-------------
+
+To load detector data that includes the range ``[start, stop]``, use::
+
+    from so3g.smurf.SmurfArchive import SmurfArchive
```
mhasself, line 32: How would you feel about:

- rename SmurfArchive.py to smurf_archive.py
- in smurf/__init__.py do "from .smurf_archive import SmurfArchive"
- Instruct users to "from so3g.smurf import SmurfArchive"

That's the style adopted in so3g.hk and so3g.proj.

mhasself (dismissed):

docs/smurf_archive.rst
jlashner: Is that something that I need to consider? Right now I think it would mess things up.

jlashner: I don't think it would actually crash, I think you would just see data from one file after another in the output, with repeated timestamps. Never tested though.

python/smurf/SmurfArchive.py
```diff
@@ -0,0 +1,300 @@
...
+    type_name = db.Column(db.String, db.ForeignKey('frame_type.type_name'))
+    frame_type = relationship('FrameType')
+
+    time = db.Column(db.DateTime, nullable=False)
+
+    # Specific to data frames
+    samples = db.Column(db.Integer)
+    channels = db.Column(db.Integer)
+    start = db.Column(db.DateTime)
+    stop = db.Column(db.DateTime)
```
ahincks, line 63: I think if you are sure that there are only metadata for data frames, then the implementation here is fine—`samples`, `channels`, `start` and `stop` are just `NULL` if the frame type is different.

If, however, you envisage eventually storing metadata from different frame types, then I would break these columns out into a separate table—something like `data_frame_metadata`—and reference it with a foreign key (and do the same for the other frame types.

python/smurf/SmurfArchive.py
```diff
@@ -0,0 +1,300 @@
...
+
+            timestamp = frame['time'].time / core.G3Units.s
+            db_frame.time = dt.datetime.fromtimestamp(timestamp)
+
+            data = frame.get('data')
+            if data is not None:
+                db_frame.samples = data.n_samples
+                db_frame.channels = len(data)
+                db_frame.start = dt.datetime.fromtimestamp(data.start.time / core.G3Units.s)
+                db_frame.stop = dt.datetime.fromtimestamp(data.stop.time / core.G3Units.s)
```
ahincks, line 146: Have you double checked that SQLite gives you enough precision in storing the time? I we don't need much sub-second precision, but we might not want to be limited to whole numbers.

python/smurf/SmurfArchive.py
```diff
@@ -0,0 +1,300 @@
...
+
+Base = declarative_base()
+Session = sessionmaker()
+
+class Files(Base):
+    """Table to store file indexing info"""
+    __tablename__ = 'files'
+    id = db.Column(db.Integer, primary_key=True)
+
+    path = db.Column(db.String, nullable=False)
```
ahincks, line 26: Would recording the MD5SUM or similar be helpful in any situation?

ahincks:
Sorry for not getting to this earlier! I've added some comments.

Oh, now I see that I can't approve. Well, in practice I'm fine with this PR being merged!

ahincks (dismissed):

jlashner:
Addressed Matthew's comments but not yet Adam's. Do you think we can merge this so people at UCSD can start using it for their analysis, and then have a new PR that implements Adam's remaining suggestions and support for multiple stream-id's?

ahincks:
I don't have any objection to merging. I'm sure this is going to be really helpful. (Though if any of my suggestions on the DB architecture, such as rearranging the `Frame` class, is implemented, you might want to consider if merging now will cause backward-incompatibility with DBs in the future.)

ahincks (approved):

docs/smurf_archive.rst
```diff
@@ -0,0 +1,59 @@
+Loading Detector Data
+======================
+
+The SmurfArchive module contains code for indexing and quickly loading detector
+data from an archive of streamed data written by the smurf-recorder OCS agent.
+A `data archive` should contain all timestream data recorded on a given system.
+The file paths within a data archive will look like::
+
+    <archive_path>/<date-code>/<stream-id>/<filename>.g3
```
mhasself, line 9: Pedantic comment: odd mixed use of _ and -.

docs/smurf_archive.rst
```diff
@@ -0,0 +1,59 @@
...
+
+For each stream-id, it is important that the time ranges of different files do
+not overlap or else data may not be returned in the correct order.
```
mhasself, line 12: Less pedantic comment: this implies that as long as your stream-ids are different, everything will be fine.  But in fact the system currently will not work if you have multiple stream-ids active simultaneously (with overlapping time ranges).

I know you're going to fix it later, but better to state the short-comings clearly in the docs until they are fixed.

mhasself (dismissed):
Approving, but I think you should address my two new comments before merge! :)

PR #72, opened 2020-10-09
Add an HKScanner to the HKArchiveScanner

BrianJKoopman:
Use the HKScanner to check schema validity of frames being processed by the HKArchiveScanner. This will provide more descriptive log messages when errors are encountered in HK files.

The motivation for this relates to https://github.com/simonsobs/ocs/pull/179. The OCS HK Aggregator was outputting invalid frames. Finding that was kind of a long road given the errors encountered. With this patch, when opening the files with invalid frames we get the descriptive error message:
```
ERROR (HKScanner): Frame does not have "block_names" entry, or it is not the same length as "blocks". (scanner.py:142 in __call__)
```

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -324,6 +324,7 @@ class HKArchiveScanner:
         self.frame_info = []
         self.counter = -1
         self.translator = so3g.hk.HKTranslator()
+        self.scanner = so3g.hk.HKScanner()
 
     def __call__(self, *args, **kw):
         return self.Process(*args, **kw)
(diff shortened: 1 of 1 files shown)


mhasself (changes requested):
I am on board with the goal here... but I don't think it's fair game to hotwire the logger verbosity to selectively elevate errors.  In level-based logging, the verbosity is supposed to be under the control of the end application -- the job of the library is to mark the severity of each message appropriately... So, why do all of our applications not propagate "errors" to the console!?  We need a better boilerplate for enabling logging-to-terminal.

Perhaps we should invest in python logging, throughout so3g, and not use g3_log.  Module-based logging is pretty nice.  Our "applications" will still need to initialize the system somehow, `logging.basicConfig(level=logging.WARNING)` or similar.  This is the sort of thing I'd want to put into a handy so3g.init_logging() function, but some might disagree.

PR #73, opened 2020-10-19
Make index_archive continue through file errors by default

jlashner:
Some old data frames taken on k2so don't have the proper format that we now require for smurf frames. This PR makes it so by default, if there is an error indexing a file, the session is rolled back but the indexing process continues. It also adds an argument `stop_at_error` which will force the `index_archive` function to stop if there is an error indexing a file.


mhasself (dismissed):
Maybe add a comment in the source about why you need this path?  Like what problem with older data leads to what Exception (apparently not an IntegrityError or RuntimeError...).

mhasself (approved):
Thanks.

PR #75, opened 2020-11-09
Move from Travis CI to GitHub Actions

### LINKED ISSUE #74: Test and deploy with github actions
Now that travis is changing its pricing model, it might be a good time to switch to github actions for tests.  I was working on using actions for wheel deployment, but we could use them for the general test case as well.

BrianJKoopman:
This moves the Travis CI pipeline over to GitHub Actions, following a similar structure to how we moved ocs/socs. Tests always get run on pushes and PRs. Docker images get built on pushes to master, like they do now. And that only happens if tests pass. Coverage is still published on all branches from either workflow.

Closes #74


BrianJKoopman:
Fixed a typo and rebased on latest `master`. Our 10k free credit allocation on Travis is close to running out, so I'll merge this by end of day unless there are objections.

PR #76, opened 2020-12-01
Read HK faster

mhasself:
Rewrite hk.get_data so it's faster.  For my test dataset from ACU, this improves read speed in "standard" mode by factor 4 (after the HKArchive is ready, anyway).

Improved the test a bit but there's probably some work to do there.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -7,6 +7,7 @@ frames.  Use the resulting HKArchive to perform random I/O over a
 sisock-style API.
 
 """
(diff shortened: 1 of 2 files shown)


mhasself:
Sorry for the noise... bad base... only 4 commits in this PR.

kmharrington (approved):
I checked that the load_range function still works and my very un-scientific "feeling" is that it loaded a lot faster than last time I ran the function (which was yesterday).

PR #78, opened 2020-12-05
Add SMuRF Status wrapper

jlashner:
This PR adds a SmurfStatus class, which interprets the status map that is in the metadata timestream, and pulls out useful attributes to make them accessible to users. It also modifies the `load_data` function so that the status at the requested start time is always returned along with the data.

The `SmurfStatus` class also provides some utility functions `readout_to_smurf` and `smurf_to_readout` which allows users to convert between readout channel index and the `(band, channel)` pair using the channel map.

### DIFF
--- a/python/smurf/smurf_archive.py
+++ b/python/smurf/smurf_archive.py
@@ -207,7 +207,7 @@ class SmurfArchive:
 
         session.close()
 
-    def load_data(self, start, end, show_pb=True, load_biases=False):
+    def load_data(self, start, end, show_pb=True, load_biases=True):
         """
         Loads smurf G3 data for a given time range. For the specified time range
         this will return a chunk of data that includes that time range.
@@ -221,11 +221,22 @@ class SmurfArchive:
 
         Returns
         --------
-            smurf_data : namedtuple
-                Returns a tuple ``SmurfData(times, data, status, biases)``.
-                If load_biases is False, ``biases`` will be None.
-                If there are no Scan frames in the time range, ``status`` will
-                be None.
+            Returns a tuple ``SmurfData(times, data, status, biases)``
+            with the following fields:
+
+                times (np.ndarray[samples]):
+                    Array of unix timestamps for loaded data
+                data (np.ndarray[channels, samples]):
+                    Array of data for each channel sending data in the
+                    specified time range. The index of the array is the readout
+                    channel number.
+                status (SmurfStatus):
+                    SmurfStatus object containing metadata info at the time of
+                    the first Scan frame in the requested interval. If there
+                    are no Scan frames in the interval, this will be None.
+                biases (optional, np.ndarray[NTES, samples]):
+                    An array containing the TES bias values.
+                    If ``load_biases`` is False, this will be None.
         """
         session = self.Session()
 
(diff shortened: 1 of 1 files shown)


python/smurf/smurf_archive.py
```diff
@@ -10,6 +10,8 @@
 from tqdm import tqdm
 import numpy as np
 import yaml
+import re
```
mhasself, line 13: I think you ended up not using re.

python/smurf/smurf_archive.py
```diff
@@ -296,5 +300,106 @@ def load_status(self, time, show_pb=False):
...
+            The A parameter of the readout filter.
+        filter_gain : Optional[float]
+            The gain of the readout filter.
+        filter_order : Optional[int]
+            The order of the readout filter.
+        filter_enabled : Optional[bool]
+            True if the readout filter is enabled.
+        downsample_factor : Optional[int]
+            Downsampling factor
+        downsample_enabled: : Optional[bool]
```
mhasself, line 341: Double colon

python/smurf/smurf_archive.py
```diff
@@ -296,5 +300,106 @@ def load_status(self, time, show_pb=False):
...
+        self.filter_gain = self.status.get(f'{filter_root}.Gain')
+        self.filter_order = self.status.get(f'{filter_root}.Order')
+        self.filter_enabled = not self.status.get('{filter_root}.Disabled')
+
+        ds_root = 'AMCc.SmurfProcessor.Downsampler'
+        self.downsample_factor = self.status.get(f'{ds_root}.Factor')
+        self.downsample_enabled = not self.status.get(f'{ds_root}.Disabled')
+
+        rtm_root = 'AMCc.FpgaTopLevel.AppTop.AppCore.RtmCryoDet'
+        self.flux_ramp_freq = self.status.get(f'{rtm_root}.RampMaxCnt')
```
mhasself, line 368: Not docstringed.  Also I am doubtful of the equivalence... "Cnt" would normally be a count, and freq propto 1/count...

We do want this in the status though!

mhasself (changes requested):
Hooray!  Excited to see this.  Lots of comments, sorry.

python/smurf/smurf_archive.py
jlashner: Added conversion found in the pysmurf repo and tested it gives a flux ramp of ~4 kHz

jlashner:
Awesome, thanks for the suggestions! Implemented with a few extra changes:
- load_data will load the status at the time of the first scan frame in the data range instead of the specified start time
- load_data returns a named tuple

python/smurf/smurf_archive.py
```diff
@@ -296,5 +300,106 @@ def load_status(self, time, show_pb=False):
...
+            self.freq_map[band] = band_center + subband_offset + channel_offset
+
+    def readout_to_smurf(self, rchan):
+        """
+        Converts from a readout channel number to (band, channel).
+        """
+        abs_smurf_chan = self.mask[rchan]
+        return abs_smurf_chan // 512, abs_smurf_chan % 512
+
+    def smurf_to_readout(self, band, chan):
```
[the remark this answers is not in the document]
jlashner: In the pysmurf read_stream_data function the mask sets all channels that aren't streaming to -1, so I'll follow that precedent .

mhasself (dismissed):
Great, thanks.  Approving with a couple additional minor notes:

- Docstrings still have some hard-coded 8/512.
- I just saw "num_bias_lines = 16" in the preamble... is that the best place and name for that?
- Docstring for load_data is now much less informative.  Insufficiently informative...
- Using "biases" is ok for variables but in docstrings it is worth specifying "TES biases".  There can be many kinds of biases applied in readout systems...

jlashner:
Good points. Added the docs back in to the load_data function and changed hardcoded instances of 8 and 512. I moved num_bias_lines away from the import statements, though I'm not sure if that's what you were getting at... I think I need another approval before merging, thanks!

mhasself (approved):

PR #79, opened 2020-12-11
Work to support unbiased map-maker

mhasself:
The only potentially annoying thing here is new cmake stuff for BLAS and cblas.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,12 +72,23 @@ file(GLOB MY_PYTHONS_SMURF
 # Provide list of libs to link against.
 target_link_libraries(so3g spt3g::core)
 
-# link openblas
-set(BLA_VENDOR OpenBLAS)
+# You probably want to select openblas, so pass -DBLA_VENDOR=OpenBLAS
 find_package(BLAS REQUIRED)
 if(BLAS_FOUND)
-    message("OpenBLAS found.")
-    target_link_libraries(so3g ${BLAS_LIBRARIES})
+  message("-- BLAS found: ${BLAS_LIBRARIES}")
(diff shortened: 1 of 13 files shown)


mhasself:
I'd like to merge this in the next day or so ... @BrianJKoopman  please let me know if the docker stuff is not ok or if you need more time to look at it.  @tskisner if you have advice on how best to check BLAS-in-cmake, please let me know.

This is not polished but I think it can be merged so we can take next steps.

BrianJKoopman:
My only comment on the Docker changes would be to run the `RUN apt install -y libopenblas-dev` before line 13 where files are `ADD`'d to the image. This would slightly speed up local builds in the case that `requirements.txt` is updated, since the cached image layer with `libopenblas-dev` already installed will be used. In the current file, that apt-get line would have to be run again.

tskisner:
As an addendum to this, concerning CMake BLAS / LAPACK, it looks like you are doing the right thing if you just need cblas.  Things get more complicated if you want to link to the standard (fortran) BLAS / LAPACK (in case cblas / clapack does not exist).  In this case you also need to check the fortran name mangling (additional underscore / capitals):

https://github.com/hpc4cmb/toast/blob/074666b5553ca845dc8249f87b8e68d16c06b0ee/CMakeLists.txt#L87

And of course to link directly to the fortran lapack, it helps to define the interfaces you are using:

https://github.com/hpc4cmb/toast/blob/master/src/libtoast/include/toast/math_lapack.hpp

mhasself:
Ok, thanks for the links... I have a suspicion I'll be back here in ~18-36 months to build in handling of new cases...  I did notice that on some systems (my arch) the cblas_ symbols are already in the blas library, but on others (simons1) you need to link in libcblas explicitly.

PR #81, opened 2020-12-23
Fix bug in SmurfArchive

jlashner:
This fixes a bug in the new Smurf status class which would cause data loading to crash when the `RampMaxCnt` was not found in the metadata stream.


mhasself:
It would be nice if comments in the code could explain why it is acceptable that / what situations would lead to RampMaxCnt not being present.

mhasself (approved):

jlashner:
Changed default flux ramp rate to `None` instead of 4 kHz, which I believe addresses Matthews concern from today's call.

mhasself (approved):

PR #82, opened 2021-01-07
Primary data loading and smurf software-based timestamps

jlashner:
This PR adds two features to the `load_data` function of the SmurfArchive:
- `load_data` now loads `primary` header fields if they exist (they may not exist in older data)
- load_data is smarter about calculating timestamps. 

Now instead of loading timestamps based on the `G3Timesample.times()` function which assumes equispaced data samples, the `get_sample_timestamps` function will return timestamps based on the timestream start time, and the UnixTime included in the packet header. This will also easily extend to loading the timing-system timestamps when we start getting that data. A TimingParadigm variable is also now returned which tells you which method was used.

### DIFF
--- a/python/smurf/smurf_archive.py
+++ b/python/smurf/smurf_archive.py
@@ -268,7 +268,7 @@ class SmurfArchive:
 
         Returns
         --------
-            Returns a tuple ``SmurfData(times, data, status, biases)``
+            Returns a tuple ``SmurfData(times, data, primary, status, biases, timing_paradigm)``
             with the following fields:
 
                 times (np.ndarray[samples]):
@@ -277,6 +277,12 @@ class SmurfArchive:
                     Array of data for each channel sending data in the
                     specified time range. The index of the array is the readout
                     channel number.
+                primary (Dict[np.ndarray]):
+                    Dict of numpy arrays holding the "primary" data included in
+                    the packet headers, including 'AveragingResetBits',
+                    'Counter0', 'Counter1', 'Counter2', 'FluxRampIncrement',
+                    'FluxRampOffset', 'FrameCounter', 'TESRelaySetting',
+                    'UnixTime'
                 status (SmurfStatus):
                     SmurfStatus object containing metadata info at the time of
                     the first Scan frame in the requested interval. If there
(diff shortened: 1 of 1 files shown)


python/smurf/smurf_archive.py
```diff
@@ -271,27 +331,36 @@ def load_data(self, start, end, show_pb=True, load_biases=True):
...
 
             if load_biases:
                 bias_order = [int(k[-2:]) for k in frame['tes_biases'].keys()]
                 biases[bias_order, cur_sample:cur_sample + nsamp] = frame['tes_biases']
+
+            # Loads primary data
+            if 'primary' in frame.keys():
+                for k, v in frame['primary'].items():
+                    if k not in primary:
+                        primary[k] = np.zeros(samples,)
```
mhasself, line 346: All the 'primary' fields are float64?

python/smurf/smurf_archive.py
```diff
@@ -271,27 +331,36 @@ def load_data(self, start, end, show_pb=True, load_biases=True):
...
+                    primary[k][cur_sample:cur_sample + nsamp] = v
+
+            ts, timing_paradigm = get_sample_timestamps(frame)
```
mhasself, line 349: So the timing_paradigm corresponding to the last frame will be what is returned.

Is it possible, in normal operation, to acquire and then read in a set of frames with this function that have different TimingParadigm values?  If so, then it is misleading to only tell the user about the last frame.  Perhaps have a TimingParadigm.Mixed or .Inconsistent to use in such cases?

python/smurf/smurf_archive.py
jlashner: Good catch, should be int64.

python/smurf/smurf_archive.py
jlashner: Yeah I was thinking about that as well and I had a hard time coming up with an instance where they would be mixed. They would have to do something like disable the timing system in the mid-operation, which I'm not sure is possible. Anyway it can't hurt to add TimingParadigm.Mixed, so I'll implement that.

jlashner:
Alright, implemented and tested. Let me know if there's anything else!

mhasself (approved):
Looks great!

PR #85, opened 2021-01-19
Update coveralls command for v3.0.0

BrianJKoopman:
v3.0.0 of coveralls-python now requires `--service` be specified for things to
work with Github Actions. See [1].

[1] - TheKevJames/coveralls-python#251


BrianJKoopman:
Before I merge this I'm going to add an updated badge for the README, as it is currently still linking to the old Travis CI page.

PR #87, opened 2021-01-27
Dockerfile: install pixell and qpoint

mhasself:
qpoint is installed from github latest (not currently pip-vailable).
pixell is in requirements.txt and installed through pip.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ FROM simonsobs/spt3g:0.3-16-g1341ea5
 # Set locale
 ENV LANG C.UTF-8
 
-# Additional system packages
-RUN apt install -y libopenblas-dev
+# Build tools needed for pixell; blas needed for so3g.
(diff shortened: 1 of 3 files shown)


BrianJKoopman (approved):
Looks fine to me. The one thing to consider is whether to tie the qpoint install to a tagged version or not.

nwhitehorn:
If it's useful at all, Sasha has integrated most (all?) of the functionality of qpoint into base spt3g_software (in the maps/ module), so you probably don't need to have both.

mhasself:
@nwhitehorn Interesting... but I am calling qpoint.QPoint(...).azel2bore, from Python, to get a quaternion that rotates my focal plane to equatorial sky as a function of time.  Is the equivalent exposed in spt3g.maps?  Scanning through it, the astrometry seems to be done with astropy, which makes it hard to get the full horizon->equatorial rotation, including boresight orientation.

qpoint has been a treat to work with, other than it not being on PyPI.

PR #88, opened 2021-02-11
Update to Ubuntu 20.04 based spt3g image

BrianJKoopman:
Also add python to python3 symlink package. In Ubuntu 20.04 systems python is always explicitly python2 or python3 in packages, so they drop `python` all together. Somewhere in so3g it uses `python` and would fail if the symlink didn't exist. I'm fine with this solution, but if someone who knows more about the build system wanted to explicitly state python3, that'd be great too.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,17 @@
 # A containerized so3g installation.
 
 # Build on spt3g base image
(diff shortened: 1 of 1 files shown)


mhasself (approved):
Oh, neat.  I was not aware of python-is-python3.

PR #90, opened 2021-03-14
Adds stream_id functionality

jlashner:
This PR adds support for multiple stream-id's to the SmurfArchive. This will allow it to be usable at Princeton and San Diego, where there are multiple slots. 

The following modifications are made:
- The 'stream_id' column is added to the Files database and the index function will add it based on the path.
- The load_data function now takes a `stream_id` argument, which allows you to load data for one or more stream_id's at a time.

I have a few questions that I'd love input on from @kmharrington, @mhasself, etc.
1. The `load_data` function now returns a dict mapping `stream-id` --> data object, instead of just the data object, where the stream_id's are specified as a keyword argument. By default, it will load all stream-id's found in the given time range. Though I think this is  the right behavior, it does change the structure of the return value so will mess up existing usage. Is this a huge deal? I really don't like changing default behavior, but Katie is working on her PR to extend the database and return AxisManager objects, so I think defaults will be changing pretty soon anyway...
2. This requires the `stream_id` column to be in the Files database, which requires a re-indexing. What's the best way to handle this? The easiest thing might be to just manually re-index existing databases after this is merged in, but there is likely a better way to do this. We only need to re-index the Files database which should take significantly less time, but I'm not sure how only change one database while keeping the other in sqlalchemy. @kmharrington do you know?

Lastly, we should think about how to merge this into Katie's branch. The actual changes to the code are pretty straightforward, so I don't think it'll be difficult to merge. If Katie's branch is near completion, we might want to just merge there to reduce the number of interface changes.

### DIFF
--- a/python/smurf/smurf_archive.py
+++ b/python/smurf/smurf_archive.py
@@ -18,6 +18,11 @@ from enum import Enum
 Base = declarative_base()
 Session = sessionmaker()
 num_bias_lines = 16
+SmurfData = namedtuple(
+    'SmurfData', 'times data primary status biases timing_paradigm'
+)
+type_key = ['Observation', 'Wiring', 'Scan']
+
 
 class Files(Base):
     """Table to store file indexing info"""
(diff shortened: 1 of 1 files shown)


mhasself:
@jlashner  Please close this if it is now obsolete.

jlashner:
If we've been running without this for three years its probably safe to close

PR #92, opened 2021-04-15
Add from Mask option for RangesMatrix

kmharrington:
@mhasself  You mentioned there was a larger PR with ranges features. I don't know if you were adding this one but it would help with creating flags.

### DIFF
--- a/python/proj/ranges.py
+++ b/python/proj/ranges.py
@@ -56,6 +56,13 @@ class RangesMatrix():
 
     @classmethod
     def from_mask(cls, mask):
+        """Take mask of any dimension and return a RangesMatrix of that size
+        """
+        if len(np.shape(mask))==1:
+            return Ranges.from_mask(mask)
+        if len(np.shape(mask))>2:
+            return cls( [cls.from_mask( mask[i] ) 
+                    for i in range(np.shape(mask)[0])] )
         return cls( [Ranges.from_mask(mask[i]) for i in range(np.shape(mask)[0])] )
 
     @property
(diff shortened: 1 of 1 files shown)


mhasself (approved):
Sorry; just saw the comment but I didn't realize it was a PR.  Looks fine, and I won't hold it up.

But it should have a docstring.

Annnnnddddd... RangesMatrix has important higher dimensional applications -- I regularly use ones with shape, e.g., (4, 1000, 123888).  It would be cool if this function checked len(shape), and recursed to RangesMatrix.from_mask until len(shape) == 2.

mhasself:
> Annnnnddddd... RangesMatrix has important higher dimensional applications -- I regularly use ones with shape, e.g., (4, 1000, 123888). It would be cool if this function checked len(shape), and recursed to RangesMatrix.from_mask until len(shape) == 2.

And the convention I adopted with .full, .zeros, .ones (PR imminent) is that if len(shape) == 1, then it returns a simple Ranges object, instead of like crashing or whatever.

kmharrington:
Does this make the dimensions as expected?

mhasself (approved):
Looks good.  I assume you tested that it at least doesn't crash on masks of 1, 2, 3d.  I'll have a unit test for the Ranges stuff soon.

Couple style notes:

(a) please leave whitespace around comparison operators such as  == and >.

(b) Note the following are equivalent:
```
 [some_func(items[i]) for i in range(len(items))]
 [some_func(item) for item in items]
```
I think the second form is to be preferred.  It does indeed work as hoped for ndarrays (iterating over the left-most dimension).

PR #95, opened 2021-08-06
docs: description of writing HK files with HKSessionHelper

mhasself:
Also other minor doc cleanup in hk.

### DIFF
--- a/demos/write_hk.py
+++ b/demos/write_hk.py
@@ -1,0 +1,64 @@
+# Note this demo is included directly from the package docs!
+#
+# This code generates a G3 file containing some telescope pointing
+# data in the "SO HK" format.  When expanding it, check the SO HK
+# format description to make sure your frame stream is compliant.
+
+import time
+import numpy as np
+
+from spt3g import core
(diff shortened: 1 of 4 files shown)


BrianJKoopman (dismissed):
This looks good to me. My one suggestion would be to maybe put the short example into something like `demos/hk_writer.py` and then in this docs file include the example with:
```
.. include:: ../demos/hk_writer.py
    :literal:
```

mhasself:
Oh, good idea @BrianJKoopman .  Let me know if any requests for bookbinder sims, @ahincks, and I will merge this week (and push to readthedocs).

ahincks:
> Oh, good idea @BrianJKoopman . Let me know if any requests for bookbinder sims, @ahincks, and I will merge this week (and push to readthedocs).

@mhasself:  No specific requests at this time apart from what we discussed on the call, which was, I believe, an example of how to write out an HK-compliant `.g3` file with a somewhat realistic structure and some fake data.

BrianJKoopman (approved):

PR #98, opened 2021-09-22
G3SuperTimestream

mhasself:
G3SuperTimestream supports compressed storage of 2d arrays, especially those carrying noisy time-ordered data.

This is in an advanced state, so posting for comment / testing.  Compression works and it's not totally awful.

This PR is a draft, for now, because:
- The approach to floats isn't right, yet.  The idea that a single "precision" can be applied to all data is good for sims but bad for real data, where individual calibration factors for each detector will introduce a different natural "precision", and any compromise precision could induce odd rounding artifacts.
- There isn't a convenient C++ interface for adding .data (making it difficult to integrate into the streamer).
- Haven't checked for memory leaks lately...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ add_library(so3g SHARED
   src/Ranges.cxx
   src/Rebundler.cxx
   src/Projection.cxx
+  src/G3SuperTimestream.cxx
   src/so_linterp.cxx
   src/exceptions.cxx
   src/array_ops.cxx
--- a/docs/cpp_objects.rst
+++ b/docs/cpp_objects.rst
@@ -1,11 +1,118 @@
 C++ Objects
 ===========
 
+Intervals
+---------
+
 The Intervals<T> objects have the following interface:
 
 .. autoclass:: so3g.IntervalsDouble
    :members:
 
+G3SuperTimestream
+-----------------
+
+G3SuperTimestream is for storing 2d arrays of data where the first
+axis corresponds to named channels and the second axis indexes time.
+The data arrays can have int32, int64, float32, or float64 data types.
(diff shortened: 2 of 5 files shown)


mhasself:
Test are failing because spt3g image doesn't have some new G3VectorTime stuff.  https://github.com/simonsobs/spt3g-docker/issues/5

BrianJKoopman:
> Test are failing because spt3g image doesn't have some new G3VectorTime stuff. [simonsobs/spt3g-docker#5](https://github.com/simonsobs/spt3g-docker/issues/5)

You have to update the docker tag here in so3g's Dockerfile to `0.3-58-g4240545`. That will get you the specific commit you mentioned in [simonsobs/spt3g-docker#5](https://github.com/simonsobs/spt3g-docker/issues/5).

These builds are automated, `0.3-79-g86198ac` is the latest build from yesterday if you want even more recent changes. You can find all available builds on [Docker Hub](https://hub.docker.com/r/simonsobs/spt3g/tags?page=1&ordering=last_updated). They should be available within 6 hours of a change to spt3g_software.

tskisner:
Just some general notes skimming through this:
- I like the structure of keeping the names as aliases to rows in the 2D data.  This is the same approach we use in toast for detector data (https://github.com/hpc4cmb/toast/blob/bf291a04aa27571e8ae798e9607c8530837f2666/src/toast/observation_data.py#L35)
- It is really great that all the compression information is stored internal to the class.  Currently in toast we have to store the individual detector compression parameters as top-level frame keys (https://github.com/hpc4cmb/toast/blob/bf291a04aa27571e8ae798e9607c8530837f2666/src/toast/spt3g/spt3g_export.py#L426-L433)

PR #99, opened 2021-11-01
Readthedocs updates

mhasself:
Main purpose of this is to address readthedocs configuration problems that developed recently.

Added a few more mocked imports, so more stuff will or can autodoc.

Review welcome but will merge in a day or so if I don't hear anything.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,7 @@
+# readthedocs works better if you pin things
+sphinx==4.2.0
+sphinx_rtd_theme==1.0.0
+sphinx-argparse==0.3.1
+
+# Required because it is used at root level in some modules.
+numpy
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,9 @@
+version: 2
+
+sphinx:
+    configuration: docs/conf.py
+
+python:
+    version: 3.7
+    install:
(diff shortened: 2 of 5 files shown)


docs/requirements.txt
```diff
@@ -0,0 +1,7 @@
+# readthedocs works better if you pin things
+sphinx==4.2.0
+sphinx_rtd_theme==1.0.0
+sphinx-argparse==0.3.1
+
+# Required because it is used at root level in some modules.
+numpy
```
BrianJKoopman, line 7: If it's going to be installed, should it be removed from the mock in `conf.py`?

Also, is this required for operation of so3g, or just for building the docs? (In other words, should it go in the main `requirements.txt`?) If yes, I don't think it's unreasonable to setup the `.readthedocs.yaml` to install both requirements files:
```
    install:
    - requirements: docs/requirements.txt
    - requirements: requirements.txt
```

mhasself: Yes, I will remove the mock, and add it to main requirements.txt too.

It is probably reasonable to include the main requirements.txt in the .readthedocs.yaml.  However, I like to keep builds as lean as possible and those modules are all explicitly mocked in the conf.py.  It may be a token gesture to efficiency to not install a bunch of unnecessary packages; but it is the gesture I want to make.

PR #101, opened 2021-11-29
Fixes to support tiled mapmaking

mhasself:
This supports tile-based mapmaking.  The corresponding changes in sotodlib are [here](https://github.com/simonsobs/sotodlib/commit/609fb42a06a809995cd91716b59d161337344d29).

The major improvements are:
- C++ acceleration of tile-based stuff like "what tiles are active"
- Generalization of domdir algorithm to well with tiles

Projectionist.assign_threads has options for which algorithm to use, but defaults to 'domdir' instead of 'simple'.

The truly breaking changes should only affect tiled mapmaking.  There are some changes in internal bookkeeping for Projectionist, and the get_active_tiles function returns a dict instead of a list.

### DIFF
--- a/demos/demo_proj1.py
+++ b/demos/demo_proj1.py
@@ -166,7 +166,7 @@ if 1:
 
 print('Compute thread assignments (OMP prep)... ', end='\n ... ')
 with Timer():
-    threads = pe.pixel_ranges(ptg, ofs, None)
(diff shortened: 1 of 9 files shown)


mhasself:
I have taken a good look at this to convince myself it won't break existing code that doesn't use tiled projectionists (which is everyone except the people who need these changes).

So I will merge this tomorrow unless someone wants to review it or otherwise hold things up.

PR #104, opened 2021-12-16
G3SuperTimestream: catch FLAC encoding errors

mhasself:
Also a test that seems to activate FLAC fails.  The failure path was a bit rare... single sample frames don't always trigger it.


jlashner (approved):
Looks good and fixes the problem I was encountering before. Thanks!

PR #105, opened 2021-12-20
Times convert to long, not int64_t

keskitalo:
This PR fixes an error that prevented compiling `G3SuperTimestream.cxx` on MacOS+MacPorts with gcc 10.3. `G3VectorTime` can automatically cast to `long` but not `int64_t`.


mhasself (approved):
Cool, thanks.  I assume there were no problems with the decompression on load, then? That constructs  G3VectorTime from vector<int64_t> iterator...

PR #107, opened 2021-12-20
No empty hits

keskitalo:
Raise an error when none of the hits fall on any of the tiles.


keskitalo:
@mhasself I reverted the `active_tiles` commit in this PR and rebased against master.

mhasself:
As discussed a long while ago, I don't think it should be a fatal error for a projector to not strike a map.  The consumer, if they expect a non-empty map, should check that.

Sorry to leave this dangling for so long.

PR #110, opened 2022-01-24
Clean up wheel build matrix and support python-3.10

tskisner:
This avoids duplication in the wheels.yml file and adds support for python 3.10.


tskisner:
@mhasself , see if you like this cleanup of the build matrix.  This also enables wheels for python 3.10.

mhasself (approved):
Lovely ... I was wondering how one does the "include" thing with matrix (I was skimming the docs and was almost there when next telecon started...)

PR #111, opened 2022-01-25
Fix typo in deploy action

tskisner:
Forgot to pip install twine after last update


tskisner:
This is a one-line fix for a typo in file that only gets used when a tag is made...  Self-merging.

PR #112, opened 2022-01-25
Fix numpy ABI constraints when building wheels

tskisner:
When building wheels, use the most recent numpy version supporting the oldest ABI version.  Temporarily enable wheel tests on push.


tskisner:
Ok, this is now ready for review:
    
* Pin numpy in our build environment to be the latest version which uses the oldest ABI supported by the python version in use.  This is 1.19 for python 3.7-3.9 and version 1.21 for python 3.10.
    
* Remove upload of sdist (source tarball) since those will not build cleanly on user systems.

* Update spt3g to latest version after upstream cmake fix.  Now get the Spt3gConfig.cmake from the spt3g install prefix.
    
* Bump bundled OpenBLAS version to 0.3.19, and tweak build to match some techniques used in homebrew recipe.
    
* Bump bundled boost version to latest 1.78.0, which gives fewer warnings when compiling with gcc-11.
    
* Move build requirements used for wheels into a separate file, so that we can pin dependency versions for the build which do not need to be pinned at run time.
    
* When pip-installing dependencies, set CC and CFLAGS so that packages being built from source use correct compilers.  For example, astropy-4 in python-3.10 environments.
    
* When building qpoint, separate build and install commands.  Also pre-execute the astropy IERS download so that we can debug any failures of that operation separately from actual qpoint issues.

wheels/install_deps_linux.sh
```diff
@@ -7,24 +7,38 @@ set -e
...
+fi
+if [ ${pyver} == "3.8" ]; then
+    numpy_ver="1.20"
+fi
+if [ ${pyver} == "3.9" ]; then
+    numpy_ver="1.20"
+fi
+if [ ${pyver} == "3.10" ]; then
+    numpy_ver="1.22"
+fi
```
BrianJKoopman, line 32: Just for my own reference, how are these versions determined?

BrianJKoopman (approved):
Looks good to me!

Just had one question inline below, but don't let it hold up merging.

wheels/install_deps_linux.sh
tskisner: This is based on looking at the numpy source:

https://github.com/numpy/numpy/blob/6077afd650a503034d0a8a5917bb9a5fa3f115fd/numpy/core/include/numpy/numpyconfig.h#L50-L66

```
#define NPY_1_7_API_VERSION 0x00000007
#define NPY_1_8_API_VERSION 0x00000008
#define NPY_1_9_API_VERSION 0x00000008
#define NPY_1_10_API_VERSION 0x00000008
#define NPY_1_11_API_VERSION 0x00000008
#define NPY_1_12_API_VERSION 0x00000008
#define NPY_1_13_API_VERSION 0x00000008
#define NPY_1_14_API_VERSION 0x00000008
#define NPY_1_15_API_VERSION 0x00000008
#define NPY_1_16_API_VERSION 0x00000008
#define NPY_1_17_API_VERSION 0x00000008
#define NPY_1_18_API_VERSION 0x00000008
#define NPY_1_19_API_VERSION 0x00000008
#define NPY_1_20_API_VERSION 0x0000000e
#define NPY_1_21_API_VERSION 0x0000000e
#define NPY_1_22_API_VERSION 0x0000000f
#define NPY_1_23_API_VERSION 0x0000000f
```

Those versions are the ones where the ABI was incremented, and so we install the latest version of numpy less than that.

PR #117, opened 2022-02-01
Only import core from spt3g when testing for existence

tskisner:
We don't need to import anything besides `spt3g.core` when checking for its availability.  The import mechanism for both the bundled and an external spt3g will globally munge the "spt3g" modules entry to point to the correct place.  The bundled copy of spt3g includes the full package.  If / when the so3g packages internally uses something besides core, we should include that in this check.


mhasself (approved):
Oh good, thanks!

PR #118, opened 2022-02-08
Bools4hk

### LINKED ISSUE #116: Support G3Bool and G3VectorBool?
This came up in conversation with @jlashner -- should we support [G3Bool](https://cmb-s4.github.io/spt3g_software/moddoc_core.html?highlight=g3bool#spt3g-core-g3bool)?

Currently we don't, and on the ocs end we ensure bools don't make it to the Aggregator as of https://github.com/simonsobs/ocs/pull/229. However, as of https://github.com/CMB-S4/spt3g_software/commit/d5b304dc24de1b2d33c7d194a8d06afb6f610572 spt3g_software supports vectors of bools with `G3VectorBool` (missing this might have been what stopped us from supporting it initially?)

I just wanted to open up the topic for wider conversation.

mhasself:
Update hk processing code in case G3TimesampleMap contains G3VectorBool at all.  Almost no change was necessary -- only get_data(raw=True) seems to care what kind of vector it is (in order to concatenate in native G3 tupes).  Regardless, I beefed up the testing code a bit.

This should close #116 as far as so3g is concerned.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # A containerized so3g installation.
 
 # Build on spt3g base image
-FROM simonsobs/spt3g:0.3-105-g959caff
+FROM simonsobs/spt3g:0.3-111-ga75e311
 
 # Set locale
 ENV LANG C.UTF-8
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -43,6 +43,7 @@ _SCHEMA_V1_BLOCK_TYPES = [
     core.G3VectorDouble,
     core.G3VectorInt,
     core.G3VectorString,
+    core.G3VectorBool,
 ]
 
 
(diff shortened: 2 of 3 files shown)


BrianJKoopman (commented):
This looks good to me, but I'm wondering why we drop the generation of older schema versions in testing.

mhasself:
If there are still schema version=0 data around, I can restore that.  My memory tells me that there aren't any version=0 data around ... but if I think about it that sort of doesn't make sense!

BrianJKoopman:
I'm not sure if there are still v0 data around, but there are still v1 data around, and it looks like `TestGetData` dropped testing a v1 file as well.

mhasself:
> I'm not sure if there are still v0 data around, but there are still v1 data around, and it looks like `TestGetData` dropped testing a v1 file as well.

Sorry for the delay ... I added back a test for v1.

BrianJKoopman (approved):
Great, looks good to me. Thanks for adding the test back!

PR #120, opened 2022-02-23
Work on fixing wheel tests

tskisner:
Try lifting astropy < 5 restriction.


tskisner:
I will remove the temporary wheel test on push after approved and before merge.

mhasself (approved):
Great, glad this went the easy way :)

PR #122, opened 2022-03-18
Load data change

jeffiuliano:
This allows one to specify a `pre_proc_dir` -- a directory where pickled `HKScannerArchive` objects can be stored -- so that they do not need to be re-created on every call.  Can be passed to `HKArchiveScanner`, to be used in the `process_file_with_cache` method (which just calls the normal `process_file` method if no `pre_proc_dir` is specified).  This is then used in the `load_range` function, which now also passes the start/end times `hksc.simple`. 

If this flag isn't used, the behavior/syntax _should_ be unchanged.

The point of this is to speed up repeated loading from the same g3 files (by using this flag).  When testing it definitely seems faster, though not by quite as much as I had expected.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -595,8 +595,12 @@ class HKArchiveScanner:
         re-processing each frame, if the corresponding file exists.  Otherwise, it saves the 
         result of that processing so it can be loaded on future calls of the fn.
         """
-        folder = str(int(start_ctime/1e5))
-        path = os.path.join( self.pre_proc_dir, folder, filename )
+        if self.pre_proc_dir is None:
+            self.process_file(filename)
+            return
+
+        folder = os.path.basename(filename)[:5]
+        path = os.path.join( self.pre_proc_dir, folder, os.path.basename(filename).replace(".g3",'') )
 
         if os.path.exists(path):
             with open(path, 'rb') as pkfl:
@@ -605,8 +609,11 @@ class HKArchiveScanner:
         else:
             hksc = HKArchiveScanner()
             hksc.process_file(filename)
-            with open(paht, 'wb') as pkfl:
-                pickle.pickle(hksc, pkfl)
+            if not os.path.exists( os.path.dirname(path) ):
+                os.umask(000)
+                os.makedirs( os.path.dirname(path), mode=0o777)
+            with open(path, 'wb') as pkfl:
+                pickle.dump(hksc, pkfl)
 
         self.providers.update(hksc.providers)
         self.field_groups += hksc.field_groups
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -437,14 +437,15 @@ class HKArchiveScanner:
     HKArchive that can be used to load data more efficiently.
 
     """
-    def __init__(self, pre_proc_dir=None):
+    def __init__(self, pre_proc_dir=None, pre_proc_mode=None):
         self.session_id = None
         self.providers = {}
         self.field_groups = []
         self.frame_info = []
         self.counter = -1
         self.translator = so3g.hk.HKTranslator()
-        self.pre_proc_dir=pre_proc_dir
+        self.pre_proc_dir = pre_proc_dir
+        self.pre_proc_mode = pre_proc_mode
 
     def __call__(self, *args, **kw):
         return self.Process(*args, **kw)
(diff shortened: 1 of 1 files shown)


python/hk/getdata.py
```diff
@@ -587,6 +589,39 @@ def process_file(self, filename, flush_after=True):
...
+        else:
+            hksc = HKArchiveScanner()
+            hksc.process_file(filename)
+            if not os.path.exists( os.path.dirname(path) ):
+                os.umask(000)
+                os.makedirs( os.path.dirname(path), mode=0o777)
+            with open(path, 'wb') as pkfl:
+                pickle.dump(hksc, pkfl)
+
+        self.providers.update(hksc.providers)
```
mhasself, line 618: Unless I forget how this works, both self.providers and hksc.providers should be empty at this point (they are emptied by the flush() call, which is performed by default at the end of each file).  So I suspect you don't need this / it is misleading to include this.

mhasself (changes requested):
Thanks for this.  It was always the intention that the scan results be saved somehow, and this does it.

I have highlighted some minor concerns, and nitpicks about formatting.  Also adding @kmharrington to review.

kmharrington:
I agree with the permissions feedback from Matthew. Does this significantly speed up your use case? I had been imagining some mechanism of just holding an HKArchiveScanner in memory between calls to `load_range`, but if this version works out then I can accept it. (@sanahabhimani  it does make me really want a non-bootstrapped version though :-D )

python/hk/getdata.py
jeffiuliano: Yeah I wasn't sure if there were cases where it would be non-empty and so needed.  You're right though, it looks like it is emptied by flush.

jeffiuliano:
> I agree with the permissions feedback from Matthew. Does this significantly speed up your use case? I had been imagining some mechanism of just holding an HKArchiveScanner in memory between calls to `load_range`, but if this version works out then I can accept it. (@sanahabhimani it does make me really want a non-bootstrapped version though :-D )

So processing the file once is pretty fast.  In principle then we could do something like you suggest and keep it in memory between calls.  It's just that I didn't think of a non-complicated way to do that which didn't require a lot of book-keeping by the load_range user (like realizing that they were accessing the same g3 file a bunch and so separately making the scanner object and passing it).  Using something in /tmp for the pre_proc_dir would be similar in effect I suppose.

python/hk/getdata.py
```diff
@@ -587,6 +589,39 @@ def process_file(self, filename, flush_after=True):
...
+        if os.path.exists(path):
+            with open(path, 'rb') as pkfl:
+                hksc = pickle.load(pkfl)
+
+        else:
+            hksc = HKArchiveScanner()
+            hksc.process_file(filename)
+            if not os.path.exists( os.path.dirname(path) ):
+                os.umask(000)
+                os.makedirs( os.path.dirname(path), mode=0o777)
```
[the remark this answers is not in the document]
jeffiuliano: yeah good point.  Switched to os.chmod and flag for permissions.  in principle each user could have their own, but it makes integrating it into some of the tools in LATR Validation easier if it can be a shared directory.

jeffiuliano:
While we have this PR for the getdata stuff, one more small change.  I ran into a weird issue where for some small time ranges, certain fields would cease to exist, so I added a non-strict mode to `load_range` that moves on rather than raising a KeyError (defaults to the normal behavior though).

python/hk/getdata.py
```diff
@@ -666,6 +709,10 @@ def load_range(start, stop, fields=None, alias=None,
         data_dir - directory where all the ctime folders are. 
                 If None, tries to use $OCS_DATA_DIR
         config - a .yaml configuration file for loading data_dir / fields / alias
+        pre_proc_dir - place to store pickled HKArchiveScanners for g3 files to speed up loading
+        pre_proc_mode - permissions (passed to os.chmod) to be used on dirs and pkl files in the pre_proc_dir. No chmod if None.
+        strict - if False, skip 
+
```
mhasself, line 715: Insufficient documentation of "strict".

Wrap the pre_proc_mode description to < 78 columns please.

python/hk/getdata.py
```diff
@@ -587,6 +590,45 @@ def process_file(self, filename, flush_after=True):
             self.flush()
 
 
+    def process_file_with_cache(self, filename):
+        """Processes file specified by ``filename`` using the process_file method above.
+
+        If self.pre_proc_dir is specified (not None), it will load pickled HKArchvieScanner objects and
```
mhasself, line 596: This still says Archvie.

mhasself: Also please wrap docstrings at ~78th column or earlier.

python/hk/getdata.py
```diff
@@ -756,7 +803,14 @@ def load_range(start, stop, fields=None, alias=None,
         if field not in all_fields:
             hk_logger.info('`{}` not in available fields, skipping'.format(field))
             continue
-        t,x = cat.simple(field)
+        try:
+            t,x = cat.simple(field, start=start_ctime, end=stop_ctime)
```
mhasself, line 807: Instead of skipping the field entirely, would it make sense to instead return empty arrays?

Seems like that is a sensible default behavior, really. (Would like to hear @kmharrington opinion on that.)

python/hk/getdata.py
```diff
@@ -756,7 +803,14 @@ def load_range(start, stop, fields=None, alias=None,
...
+        except Exception as e:
+            if not strict and isinstance(e, KeyError):
+                hk_logger.error(f'{e} -- skipping field')
```
mhasself, line 810: .warning instead of .error, I think.

mhasself (changes requested):
Thanks  -- a few more comments (sorry to drag this out).

python/hk/getdata.py
jeffiuliano: Since it returns a dict ultimately rather than a list, it's not so different in use-case.  I would have a slight preference for assuming that data exists when I see a key in the returned data, but I think either would be reasonable behavior.

jeffiuliano:
Sorry it took a while to get back to this, but I made the suggested changes other than the default behavior when a field is missing.  Happy to change that too if we decide that's what we want.

mhasself (approved):
Had a chance to give this another whirl today -- definitely will speed things up in certain cases.

I think this isn't the end of the story but I'd definitely recommend folks working with long stretches of HK data to use this.

PR #132, opened 2022-10-21
Speed up HK loads

mhasself:
The low level code to process HK frames wasn't actually bothering to record, and avoid, frames that don't overlap with the requested time range.  Now it does, which should speed things up (especially in cases where only short time intervals of data -- shorter than the "load_ranges" buffer of ~1 hour -- are requested).

This will cause the output of the HKArchive data retrieval functions to change (.simple, .get_data) -- should be a subset of the data returned before these changes.  However I think most "users" rely on load_range, which does its own timestamp trimming.

Note that the internal data structure of "_FieldGroups" has changed, which means that cached HKArchiveScanner output (using pre_proc_dir) will need to be regenerated.  (Just delete all existing pickle files so they will regenerate.)  If you get errors like this:
```
get_data: _get_groups returns 1 groups.
Traceback (most recent call last):
...
KeyError: 'time_range'
```
Then it's due to this _FieldGroups change.


mhasself:
@jeffiuliano @skhrg  Please note the warning about this breaking any .pik you've saved in pre_proc_dir.

mhasself:
The impact of this PR is actually negligible once #133 is incorporated (at least on my test csae)  Note I'll soon be rebasing this on #133.

mhasself:
I am planning to merge this tomorrow.

PR #133, opened 2022-10-21
Performance improvement by avoiding python for loop

dixilo:
Previous code has Python for loop, which is consuming a significant part of processing time.
I avoided the for loop in 't' array just by using buffer pass-through from G3 like other data fields.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -327,7 +327,7 @@ class HKArchive:
                     if group_name not in timelines:
                         # This block is init that happens only once per group.
                         timelines[group_name] = {
-                            't': np.zeros(gi['count']),
+                            't_g3': np.zeros(gi['count'], dtype=np.int64),
                             'fields': [f for f,s in gi['fields']],
                         }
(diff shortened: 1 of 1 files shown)


python/hk/getdata.py
```diff
@@ -370,7 +370,8 @@ def get_data(self, field=None, start=None, end=None, min_stride=None,
 
         # Scale out time units and mark last time.
         for timeline in timelines.values():
-            timeline['t'] /= core.G3Units.seconds
+            timeline['t'] = timeline['t_g3']/core.G3Units.seconds
+            del timeline['t_g3']
```
mhasself, line 374: Please change to
```
timeline['t'] = timeline.pop('t_g3') / core.G3Units.seconds
```

mhasself (changes requested):
Excellent analysis!  These buffer-based accelerations weren't available when the so3.hk code was originally written.

This has a *huge* impact on load times, the HK world will thank you!

Just a minor change requested.

mhasself (approved):

PR #134, opened 2022-10-21
Load ranges onepass

mhasself:
(Note I'm opening _this_ PR into #132, because it's a bit more contentious.)  

@kmharrington  There are some savings to be had in load_ranges -- a single call to "simple()" will reduce the number of passes through the frame data.  The only difference in functionality here is that it doesn't catch "single field" failures.  Was there a specific failure mode you were trapping there that we could try to address some other way?

In the hk example we've been looking at, this fix increase load time by factor 2.  The improvement should scale (though not linearly, probably) with the number of fields you're trying to load.


kmharrington (approved):
looks good

PR #136, opened 2022-10-26
HKTree -- tab completion for browsing HK data

mhasself:
When trying to inspect some SAT data, I found I wanted a system where I could delve into the HK data hierarchy more easily, to see what was there and read in what I found.  This is an attempt to do that, using python/jupyter tab completion to assist with exploration.

I'd appreciate feedback on the interface (@kmharrington @BrianJKoopman @jlashner @jseibert575) in addition to a review (volunteers).  It is awkward to use private members (e.g. `_load`) to do things but I think it's worth it to be able to use tab completion and concise hierarchical field names.

**Preamble from the documentation:**

The HKTree object provides a way to browse through available HK data fields from within an interactive python session (such as through the python or ipython interpreters or in a jupyter session).

Instead of accessing HK fields through their long string names, such as:
```
field_name = "observatory.hk.rotation.feeds.hwprotation.kikusui_curr"
```
a field can be referred to as a named object in a hierarchy of attributes:
```
tree = HKTree('2022-10-20', '2022-10-22', config='hk.yaml')
field = tree.hk.rotation.hwprotation.kikusui_curr
```
By exposing the available fields as a tree of attributes, rather than huge set of long string keys, a user working interactively can use tab-completion to find fields easily.

Having identified a field or set of fields of interest, the data can be loaded by calling pseudo-private methods on the field directly, e.g.:
```
data_dict = field._load()
```
After calling load, the data are stored in the field reference itself, and so can be retrieved with:
```
times, values = field._data
```
The _load method can be called on “non-terminal” nodes of the tree, for example:
```
data_dict = tree.hk.rotation._load()
```
would load all the fields that are sub- (or sub-sub-, …) attributes of tree.hk.rotation.

The system is intended to complement so3g.hk.load_range, and uses the same sort of configuration file.

### DIFF
--- a/docs/hk.rst
+++ b/docs/hk.rst
@@ -144,6 +144,174 @@ Function References
 .. autofunction:: so3g.hk.getdata.to_timestamp
 
 
+Exploring HK Data
+-----------------
+
+The HKTree object provides a way to browse through available HK data
+fields from within an interactive python session (such as through the
+python or ipython interpreters or in a jupyter session).
+
+Instead of accessing HK fields through their long string names, such as::
+
+  field_name = "observatory.hk.rotation.feeds.hwprotation.kikusui_curr"
+
+a field can be referred to as a named object in a hierarchy of
+attributes::
+
+  tree = HKTree('2022-10-20', '2022-10-22', config='hk.yaml')
+  field = tree.hk.rotation.hwprotation.kikusui_curr
+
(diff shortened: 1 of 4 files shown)


mhasself:
I'm planning to add some basic tests for this and then merge.  If anyone wants to review, please let me know before the weekend.

PR #137, opened 2022-10-31
Changes the to_timestamp function to work with np  types

jlashner:
Small change to the getdata module that makes the `to_timestamp` function work with numpy datatypes. If a number was of type `np.float64` checking equality of its type against float fails, but the `isinstance` function works as expected.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -686,7 +686,7 @@ def to_timestamp(some_time, str_format=None):
     
     if type(some_time) == dt.datetime:
         return some_time.astimezone(dt.timezone.utc).timestamp()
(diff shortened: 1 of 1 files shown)


kmharrington (changes requested):
If we're going to edit this function could we use something like what we built in sotodlib. That on is much more general: https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/load_smurf.py#L208

mhasself:
@kmharrington To be clear, is it enough for @jlashner to use the "np.issubtype" logic that the first line of the sotodlib func uses?

mhasself:
Closing as abandoned.

PR #138, opened 2022-11-09
Update wheel building scripts

tskisner:
* Use same pre-built openblas binary that ships with numpy wheels.

* Add support for python-3.11

* Build with clang on MacOS

* Fix readme

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,6 +38,10 @@ jobs:
             python: 310
             platform_id: manylinux_x86_64
             manylinux_image: manylinux2014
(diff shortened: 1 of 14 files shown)


Dobatymo:
`pysqlite3-wheels` is now available for newer python versions. So you don't have to remove it.

tskisner:
The wheels are now building macos for python-3.9 to 3.11.  You can see the full list here:

https://github.com/simonsobs/so3g/actions/runs/4465041253?pr=138

I will now switch the tests back to their normal configuration.

mhasself (approved):
This is all terrifying, thanks for taking care of it!

PR #139, opened 2022-11-19
so-hk-cli

mhasself:
Tool for scanning HK files; see included docs.

This might need some additional work, in setup.py, to get installed properly when using the pip method -- would appreciate @tskisner advice.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,7 @@ install(FILES ${MY_PYTHONS_PROJ}
 install(FILES ${MY_PYTHONS_SMURF}
         DESTINATION ${INSTALL_DEST}/smurf)
 
+# To add a prefix, pass CMAKE_INSTALL_PREFIX.
 install(PROGRAMS scripts/so-hk-tool DESTINATION bin)
 
 add_custom_target(prep-readthedocs
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,8 @@ install(FILES ${MY_PYTHONS_PROJ}
 install(FILES ${MY_PYTHONS_SMURF}
         DESTINATION ${INSTALL_DEST}/smurf)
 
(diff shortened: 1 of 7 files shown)


jlashner:
I'm trying to run `make install` and it seems like it's trying to install this to `/usr/local/bin`. Is there an easy way to install this to `~/.local/bin` instead?

```
-- Up-to-date: /home/jlashner/.local/lib/python3.7/site-packages/so3g/smurf/reader.py
-- Up-to-date: /home/jlashner/.local/lib/python3.7/site-packages/so3g/smurf/smurf_archive.py
-- Installing: /usr/local/bin/so-hk-tool
CMake Error at cmake_install.cmake:155 (file):
  file INSTALL cannot copy file
  "/home/jlashner/repos/so3g/scripts/so-hk-tool" to
  "/usr/local/bin/so-hk-tool": Permission denied.
```

CMakeLists.txt
```diff
@@ -142,6 +142,8 @@ install(FILES ${MY_PYTHONS_PROJ}
 install(FILES ${MY_PYTHONS_SMURF}
         DESTINATION ${INSTALL_DEST}/smurf)
 
+install(PROGRAMS scripts/so-hk-tool DESTINATION bin)
```
jlashner, line 145: Can we change this to `${CMAKE_INSTALL_BINDIR}`

jlashner (dismissed):
Works great for me! My one request would be to have an option that makes it print out sizes in a human-readable format, since it's currently pretty difficult to large sizes by eye.

CMakeLists.txt
mhasself: My version of cmake doesn't know about that ... added some notes to docs about using CMAKE_INSTALL_PREFIX to change this.

mhasself:
Thanks @jlashner .

Force-pushed to clear out hk-tree stuff (sry).  Also added units (e.g. -BM for megabytes), and a sort option.

PR #141, opened 2022-12-20
Only instance of deprecated data types

keskitalo:
This line is causing CI failures in `sotodlib`. See for example https://github.com/simonsobs/sotodlib/actions/runs/3743606379/jobs/6355965635#step:8:142


mhasself (approved):
Thanks!  I'm surprised this block works at all ... a separate problem, though.

PR #150, opened 2023-07-18
Handle empty simple hk

mhasself:
In cases where a time range did not have any data for a requested field, .get_data simply does not include that result, and .simple() was then failing to construct the return list, which was supposed to have the same length as the list of requested fields.

Test case:
```
clean_start = 1628726418
clean_stop = 1628726438

import so3g
from so3g import hk
data_dir = '/mnt/so1/data/ucsd-sat1/hk'

data = so3g.hk.load_range(start=clean_start, stop=clean_stop, data_dir=data_dir)
```

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


kmharrington (dismissed):
Thanks for taking a look!

PR #152, opened 2023-08-16
Promote qpoint and pixell to be full dependencies

mhasself:


tskisner (approved):
I think that these packages are genuine dependencies (i.e. we do use and need them), so it is good to have them listed.  I see the last `qpoint` wheel was from more than a year ago and there are no wheels for python-3.11.  But that is just a matter of pinging the developers if it is a hold-up.

arahlin:
> I think that these packages are genuine dependencies (i.e. we do use and need them), so it is good to have them listed. I see the last `qpoint` wheel was from more than a year ago and there are no wheels for python-3.11. But that is just a matter of pinging the developers if it is a hold-up.

Developers pinged.  I've got a new version uploading now that includes wheels for python-3.11 and 3.12.

Issue #151, opened 2023-08-03
HK logger does too much logging


newburgh:
Katie: anyone know where this logger is coming from and how I can shut it up?

![image](https://github.com/simonsobs/ocs/assets/7266917/d16486d8-55fe-4106-b34f-5cf6482b3e6e)

mhasself:
This is only noisy when running through Prefect, as I understand it.

We can turn down the verbosity in so3g but a better understanding of Prefect x Python logging is probably on the horizon.

guanyilun:
prefect is logging at INFO level I think. Maybe this needs to be at DEBUG level? https://github.com/simonsobs/so3g/blob/master/python/hk/getdata.py#L315

mhasself:
Resolved in #153

PR #153, opened 2023-08-17
Suppress hk logger.info messages

mhasself:
These (esp the getdata fileloop) are spamming prefect logs. Reprioritize when someone understand python logging better ...


mhasself:
(Addresses #151 )

kmharrington (approved):

PR #155, opened 2023-08-17
Add set of optional test requirements.

tskisner:


mhasself (approved):
Thanks for following this up!

PR #156, opened 2023-08-23
Add ability for load_range() to read HK books

sanahabhimani:
`load_range()` assumes the path structure is `/hk/<timecode>` but the books are
`hk/hk_16902_satp3`,  `hk_16903_satp3`, for example.  

I added an argument to `load_range()` called `platform` which you define as `'satp2'`, `'lat'`, `'site'`, etc. This triggers `load_range()` to know we're dealing with the books path structure.

Has been tested on HK books at the site.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -772,7 +772,7 @@ def load_range(start, stop, fields=None, alias=None,
         files to speed up loading
       pre_proc_mode: Permissions (passed to os.chmod) to be used on
         dirs and pkl files in the pre_proc_dir. No chmod if None.
-      platform: String of type of HK book (satp1, lat, site) to load,
+      platform: String of type of HK book (Ex: satp1, lat, site) to load,
         required to open books as book directories are more specific.
       strict: If False, log and skip missing fields rather than
         raising a KeyError.
@@ -847,7 +847,7 @@ def load_range(start, stop, fields=None, alias=None,
     for folder in range( int(start_ctime/1e5), int(stop_ctime/1e5)+1):
         if platform is not None:
             book_path = 'hk_'+str(folder)+'_'+platform
-            base = '/'+data_dir+'/'+str(book_path)
+            base = data_dir+'/'+str(book_path)
         else:
             base = data_dir+'/'+str(folder)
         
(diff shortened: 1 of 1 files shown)


python/hk/getdata.py
```diff
@@ -836,19 +838,27 @@ def load_range(start, stop, fields=None, alias=None,
...
     stop_ctime = to_timestamp(stop) + 3600
 
     hksc = HKArchiveScanner(pre_proc_dir=pre_proc_dir)
-    
+
     for folder in range( int(start_ctime/1e5), int(stop_ctime/1e5)+1):
-        base = data_dir+'/'+str(folder)
+        if platform is not None:
+            book_path = 'hk_'+str(folder)+'_'+platform
+            base = data_dir+'/'+str(book_path)
```
kmharrington, line 850: I'm glad this fix looks as simple as I was hoping. We might be able to make it even simpler? I'm not sure we need to have platform actively specified. Instead we could just look for the pattern `hk_ctime_{something}`? 

The folder structure for the site will be preserved for the librarian, so books for each `daq-node` will always be in the same folder.

sanahabhimani: Great! That was basically my question last week; if there's a way to generalize this even further. My hunch was that I can likely assume the path to hk books will always be `/so/blah/satp1/hk` or `/so/blah/site/hk`, so you can use that path without needing the `platform` arg. Is that what you were thinking?

I didn't end up doing that outright because I first tested on E2E books which didn't have a path of this structure to get to the ocs3 HK books and so the `platform` arg was useful there. But assuming people who explore HK books will never copy any books to their specific directories and preserve some form `/so/blah/satp1/hk`, we don't need the  `platform` arg

kmharrington: you could leave it at optional in case folks want to specify it. But have it so if platform=None and the code encounters an `hk_ctime_somthing` path it just runs with it. 

Side note, what you're calling platform we've been calling daq_node in the data packaging, because there's also site.

sanahabhimani:
Okay, expanded and reworded to replace `platform` arg to `daq_node` and now `daq_node` is optional and it checks to see if the node can be defined from `data_dir`

kmharrington (approved):
Ok, one thing to change and then add it. 

Use the `hk_logger` instead of print statements. That helps make sure the print statements go where we want. Also, does this now make a print out every time we want to load data from older level 2 data (like on simons1)? That might be annoying, so I might say this should be a `hk_logger.debug` statement.

kmharrington (approved):

PR #161, opened 2023-12-20
Bilin tweaks

mhasself:
Updates the description of "threads" argument.  Update tests for broader support of threads, and instantiate interpolated projectors directly.  Fix a bug where invalid RangesMatrix could be constructed and was being constructed.

This breaks current sotodlib usage in some cases (projections with threads and cuts), so don't merge until a couple of days after the related [sotodlib fix](https://github.com/simonsobs/sotodlib/pull/639).


amaurea (approved):
Looks good

PR #162, opened 2023-12-27
Load range level3 Fix

sanahabhimani:
fixed the ability for` load_range` to read both books and .g3 files. testing on satp3 level 2 and level 3 data using config file, alias/fields, and fields arguments on the site setup.

### DIFF
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -851,23 +851,27 @@ def load_range(start, stop, fields=None, alias=None,
             node = i
     
     for folder in range( int(start_ctime/1e5), int(stop_ctime/1e5)+1):
-        if daq_node is None:
-            if node in data_dir:
-                book_path = 'hk_'+str(folder)+'_'+node
-                base = data_dir+'/'+str(book_path)
-            else:
-                hk_logger.debug(f'No daq node info provided in {data_dir}, and'
-                                'daq_node arg is None; going to assume data_dir'
-                                'points to .g3 files')
-                # assumes .g3 files but should be more explicit
-                base = data_dir+'/'+str(folder)
+        base = os.path.join(data_dir, str(folder))
+        if not os.path.exists(base):
+            # see if book exists instead 
+            #hk_logger.debug(f'{base} does not exist')
+            if daq_node is None:
+                if node in data_dir:
+                    book_path = 'hk_'+str(folder)+'_'+node
+                    base = data_dir+'/'+str(book_path)
+                else:
+                    hk_logger.debug(f'No daq node info provided in {data_dir}, and'
+                                    'daq_node arg is None; going to assume data_dir'
+                                    'points to .g3 files')
+                    # assumes .g3 files but should be more explicit
+                    base = data_dir+'/'+str(folder)
         else:
             book_path = 'hk_'+str(folder)+'_'+daq_node
             base = data_dir+'/'+str(book_path)
         
-        if not os.path.exists(base):
-            hk_logger.debug('{} does not exist, skipping'.format(base))
-            continue
+        #if not os.path.exists(base):
+        #    hk_logger.debug('{} does not exist, skipping'.format(base))
+        #    continue
     
         for file in sorted(os.listdir(base)):
             for i in file.split('.'):
--- a/python/hk/getdata.py
+++ b/python/hk/getdata.py
@@ -854,25 +854,16 @@ def load_range(start, stop, fields=None, alias=None,
         base = os.path.join(data_dir, str(folder))
         if not os.path.exists(base):
             # see if book exists instead 
-            #hk_logger.debug(f'{base} does not exist')
             if daq_node is None:
                 if node in data_dir:
                     book_path = 'hk_'+str(folder)+'_'+node
                     base = data_dir+'/'+str(book_path)
                 else:
-                    hk_logger.debug(f'No daq node info provided in {data_dir}, and'
-                                    'daq_node arg is None; going to assume data_dir'
-                                    'points to .g3 files')
-                    # assumes .g3 files but should be more explicit
+                    hk_logger.debug(f'No daq node info provided in {data_dir}, cannot'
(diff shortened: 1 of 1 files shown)


python/hk/getdata.py
```diff
@@ -851,28 +851,22 @@ def load_range(start, stop, fields=None, alias=None,
...
+                    hk_logger.debug(f'No daq node info provided in {data_dir}, cannot'
+                                    'find HK book. Assuming path is to .g3 files.')
+                    # assumes path is to .g3 files instead of HK book
+                    base = data_dir+'/'+str(folder)
+        
         for file in sorted(os.listdir(base)):
-            for i in file.split('.'):
-                if i == '.yaml':
-                    continue
+            if i in file.split('.') == '.yaml':
```
mhasself, line 868: Just lurking but I think you mean
```
if 'yaml' in file.split('.'):
```
Are you actually looking for this though?: `if file.endswith('.yaml')`.

mhasself (changes requested):
Thanks for the fix attempt here!  I think there a few problems with the daq_node / node variables that mean this needs to be revised further.

I missed the earlier PR that introduced these changes, so forgive me for a few comments that go beyond the scope of the present fixes.

Please to the docstring of load_range a brief description of the two possibilities for the directory structure; either simple 5-digit folders vs. book-style naming for a specific platform.

Revise the documentation of the "daq_node"; it currently has a weird second sentence.

The "daq_node" argument will not be processed properly, if daq_node != None.  It looks like daq_node and node are sharing a purpose here, and that needs to be resolved more carefully.  I think you can do everything you want with just daq_node (as a function parameter that defaults to None), which gets upgraded based on the data_dir if need be.

On the variable "node" -- the code tries to guess a good value for node by looking at the data_dir.  If that fails, node remains _undefined_, which is dangerous.  Instead, you should set node = None, then search for a better value.  But basically I think there's just a degeneracy with daq_node as described above.

In the folder loop, there's a test for `if node in data_dir`.  Because of the reasons above, this will either fail (node not defined) or definitely be true (because node was set based on testing a limited set of strings against data_dir.

Have you considered simply testing, for each iteration of folder, whether _any_ of the following patterned folders exist?:
- {folder}  (e.g. 17890)
- *_{folder}_* (e.g. hk_17890_satp1, or whatever_17890_blarg, etc)

This would make the code less rigid -- we wouldn't have to hard-code the list of valid platforms into the function.

sanahabhimani:
the docstring could be cleaner for folder_patterns arg but ready for re-review overall.

python/hk/getdata.py
```diff
@@ -841,14 +848,34 @@ def load_range(start, stop, fields=None, alias=None,
...
+
+    if folder_patterns is None:
+        folder_patterns = ['{folder}', 'hk_{folder}_*']
     for folder in range( int(start_ctime/1e5), int(stop_ctime/1e5)+1):
-        base = data_dir+'/'+str(folder)
-        if not os.path.exists(base):
-            hk_logger.debug('{} does not exist, skipping'.format(base))
+        bases = []
+        for pattern in folder_patterns:
+            extended_pattern = pattern.replace('{folder}', str(folder))
```
mhasself, line 858: This is a bit unusual.  One would normally do
```
extended_pattern = pattern.format(folder=folder)
```

python/hk/getdata.py
```diff
@@ -841,14 +848,34 @@ def load_range(start, stop, fields=None, alias=None,
...
+            
+            base = glob.glob(os.path.join(data_dir, extended_pattern))
+            bases.append(base)
```
mhasself, line 861: `bases.extend(base)`

python/hk/getdata.py
```diff
@@ -841,14 +848,34 @@ def load_range(start, stop, fields=None, alias=None,
...
+        
+        # remove empty lists in bases
+        bases = list(filter(None, bases))
```
mhasself, line 864: Remove this.

python/hk/getdata.py
```diff
@@ -841,14 +848,34 @@ def load_range(start, stop, fields=None, alias=None,
...
+        
+        if len(bases) > 1:
+            bases.sort
+            base = bases[0][0]
+            hk_logger.warn(f"Multiple base folders were found for {folder}. The first one, alphabetically, is selected: {base}")
+        elif len(bases) == 1:
+            base = bases[0][0]
```
mhasself, line 871: `base = bases[0]`

python/hk/getdata.py
```diff
@@ -804,6 +809,8 @@ def load_range(start, stop, fields=None, alias=None,
             'HAN 1', 'HAN 2',
         ]
 
+        folder_patterns = ['hk_{folder}_satp3']
```
mhasself, line 812: This isn't used below.

python/hk/getdata.py
```diff
@@ -772,6 +773,10 @@ def load_range(start, stop, fields=None, alias=None,
         files to speed up loading
       pre_proc_mode: Permissions (passed to os.chmod) to be used on
         dirs and pkl files in the pre_proc_dir. No chmod if None.
+      folder_patterns:  List of patterns to search for in folders. If
+        None, default pattern is ['{folder}', 'hk_{folder}_*']. If not
+        None, usage for .g3 folder: ['{folder}'], and example usage 
+        for HK books: ['hk_{folder}_lat']
```
mhasself, line 779: Mention that "{folder}" will be replaced with the first 5 digits of the unix timestamp.  Or something like that.

sanahabhimani:
Yes, can confirm that I have tested on level2 and level3 for satp3. Mostly using config file arg, with and without folder_patterns arg using .g3 files and HK books. Also made a new directory in my home dir that had a mix of HK satp3 books and satp3 .g3 files (though didn't add HK books from other telescopes..) and tested with that.

kmharrington (approved):
Looking like this will work out! Thanks!

Issue #147, opened 2023-04-02
so3g.proj: det_weights=0 isn't optimized away


mhasself:
One can pass in custom cuts to avoid projecting dets that aren't of interest, but we should support short-circuits based on det_weights==0 too.

tterasaki:
I am working in the branch of [single_det_map_modification](https://github.com/simonsobs/so3g/blob/a6341c405a7a758149e12f600bcb2f678089fa1c/src/Projection.cxx#L1393).

I have just added
```
if (det_wt == 0.) continue;
```
in the `to_map_single_thread`, `to_weight_map_single_thread`, `precomp_to_map_single_thread`, and `precomp_to_weight_map_single_thread`.

I am demonstrating it in `compute11:/so/home/tterasaki/code_dev/so3g_proj_singledetmaps_demo.ipynb`
(Actually I am not working in the compute11 as it was difficult to compile so3g in compute11. Its just for sharing)

The newer version works ~factor 2 faster when n_det is smaller than 100. Howerver when n_det is about 500 or larger, it gets super slow and slower than the 'restrict and makemap' method. I am not so familiar with c++, but I thought for-loop is not so bad operation in c++. So I do not understand why it is so slow. Do you have any idea, @mhasself ?

mhasself:
Welcome to C / C++!  Indentation doesn't matter in C / C++.  Do have multiple statements in an "if" or "else" (or anything) block, surround them in {}.

So in C, this:
```
    if (something)
        thing1 = 2;
        thing2 = 9;
```
is the same as
```
    if (something)
        thing1 = 2;
    thing2 = 9;
```

Multiple statements in an if block need to be inside {}, like this:
```
    if (something) {
        thing1 = 2;
        thing2 = 9;
    }
```

mhasself:
However, I don't see why that minor formatting issue would matter here (it's ok for the check to be outside of the if block).

I'm not sure why it would matter, but maybe the thread assignment is an issue?  What happens if you create P with threads=False?

tterasaki:
Thanks Matthew! I was obsessed with python style before I know... Will fix later.

Without fixing the formatting issue, I tried the `threads=False` case by comparing the following two.
v2 is `threads=None` and v3 is `threads=False`. 

```Python
def make_singledetmaps_v2(aman, res=0.3*coords.DEG):
    wcs_kernel = coords.get_wcs_kernel('car', 0, 0, res)
[...]

PR #165, opened 2024-01-17
Optimization of the proj for the case of some of det_weights == 0.0

tterasaki:
This PR makes the map making faster in the case of some of det_weights == 0.
Applications are making single_det_maps for all detector or managing null splits.
For the detail, follow conversation in issue #147 .


mhasself (approved):
Thanks for implementing and testing!

PR #167, opened 2024-02-06
Bump spt3g version

tskisner:
Testing build


mhasself:
@tskisner If you could cast your eyes on #173 and #174 I'd appreciate it.  Then we'll be in a bit of a holding pattern before we can bump the spt3g version ... but I would like to do that as soon as it is possible (see #175).

tskisner:
Superceded by #174 and #177

PR #168, opened 2024-03-13
RangesMatrix: __getindex__ supports advanced indexing

mhasself:
Just like numpy arrays (hopefully), except that last axis is treated a bit specially (which is not new).  Extensive tests.

This came up in @kmharrington testing AxisManager wrapping / concatenation operations involving 3-dimensional RangesMatrix.


kmharrington (approved):
Thanks!

PR #175, opened 2024-04-05
Test new spt3g

mhasself:
Build + test with bumped spt3g version, plus dload change in #174

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@
 # A containerized so3g installation.
 
 # Build on spt3g base image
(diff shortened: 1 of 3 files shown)


mhasself:
Just testing a build + test with all the recent changes _and_ a version of spt3g that has dload.

When it comes to bumping spt3g version I'd like to wait for new "spt3g" docker image that has ubuntu 22.04 (https://github.com/simonsobs/spt3g-docker/pull/6, after which we'll need spt3g to commit something).

PR #179, opened 2024-05-16
Healpix mapmaker

earosenberg:


demos/demo_proj_hp.py
```diff
@@ -0,0 +1,168 @@
...
+    idx = (weights != 0)
+    iweights[idx] = 1/(weights[idx])
+    imap = (rhs * iweights[0])[0]
+
+    ## Make a tiled map ##
+    # Tiles used for thread assignment and to efficiently store partial-sky information. Much better performance / thread scaling than untiled
+    # You want a few tiles per thread to allow load balancing so set this for nActiveTiles/nthreads = 12*nside_tile**2 * fsky / nthreads ~5-10
+    print("Making tiled map")
+    proj = so3g.proj.ProjectionistHealpix.for_healpix(nside, nside_tile=nside_tile)
+    proj.assign_threads(asm, 'tiles') ## Assign threads
```
mhasself, line 57: To test threading, you have to capture the return value
```
  threads = proj.assign_threads(...)
```
and pass it explicitly to to_map / to_weights.  (I tried this, it works well.)

demos/demo_proj_hp.py
```diff
@@ -0,0 +1,168 @@
...
+    proj.assign_threads(asm, 'simple') ## Assign threads
+    rhs = proj.to_map(signal, asm, det_weights=det_weights, comps='T')
+    weights = proj.to_weights(asm, det_weights=det_weights, comps='T')
+    # Now simple combination
+    iweights = np.zeros_like(weights)
+    idx = (weights != 0)
+    iweights[idx] = 1/(weights[idx])
+    imap = (rhs * iweights[0])[0]
+
+    ## Make a tiled map ##
```
mhasself, line 52: The demo (or ideally tests) should include testing from_map, too.

My naive attempt:
```
        proj = so3g.proj.ProjectionistHealpix.for_healpix(nside, nside_tile=None)
        signal2 = proj.from_map(rhs, asm)
```
Caused to a segfault.  So that needs to be fixed.  If implementing `from_map` is problematic for some reason, then at least protect the user from segfault with appropriate argument checks.

include/healpix_bare.h
```diff
@@ -0,0 +1,118 @@
+/* -----------------------------------------------------------------------------
+ *
+ *  Copyright (C) 1997-2019 Krzysztof M. Gorski, Eric Hivon, Martin Reinecke,
+ *                          Benjamin D. Wandelt, Anthony J. Banday,
+ *                          Matthias Bartelmann,
+ *                          Reza Ansari & Kenneth M. Ganga
+ *
+ *  Public interface of the Healpix bare bones C library
+ *
+ *  Licensed under a 3-clause BSD style license - see LICENSE
```
mhasself, line 10: In the so3g LICENSE file, append a section like
```
healpix_bare.h and healpix_bare.c are included in this source repository under the terms of the following license:
...
```
and include the license from healpix.

mhasself (changes requested):
Sorry it took me a while to get to this.  This is appreciated!

A few comments inline.  Also:
- The demo is helpful, but there should be unit tests too (in test/).  (See test_proj_eng.py ... I think it's best if unit tests run quickly, i.e. using very small problems for the trials.)
- Add some basic instructions to the docs (doesn't have to be exhaustive; just a starting point / example).

earosenberg:
Thanks for the detailed comments! With respect to the docs, my restructuring of the `Projectionist` class means that the automatic reference for `Projectionist` is now missing the shared methods belonging to the parent class. I could just add all three classes (`_ProjectionistBase`, `Projectionist`, `ProjectionistHealpix`) to the automatic docs, but don't know if there's a preferred solution?

earosenberg:
I believe these commits address all of the above points, except for the docs issue I mentioned in my previous comment

mhasself (changes requested):
Thanks for the additional fixes.  This is almost ready.

For the Projectionist docs, I think you just need to put all 3 classes in the autodocs, and then create an explicit link in the docstring for the two subclasses saying "see also the methods defined in base class, ...".  I think it's dangerous and annoying to duplicate docstring text ... but helpful to provide clickable links in the rendered documentation to the one source of truth.  (This approach is already in there, where method docstrings don't _all_ explain what is meant by arguments "assembly", "comps", etc ... but rather that's explained in the class docstring.  All that stuff in the Projectionist docstring should probably go in the base class docstring, then, and just a link remains (along with everything specific to RectPix).)

python/proj/coords.py
```diff
@@ -125,7 +125,7 @@ def naive_az_el(cls, t, az, el, roll=0., site=None, weather=None):
         return self
 
     @classmethod
-    def az_el(cls, t, az, el, roll=None, site=None, weather=None):
+    def az_el(cls, t, az, el, roll=None, site=None, weather=None, **kwargs):
```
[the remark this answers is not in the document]
earosenberg: Not directly; it was added because extra qpoint args were necessary to get a good match to Toast sims in my validation tests. Can remove from this PR if you want, else I'll add a note to the docstring like you say

python/proj/wcs.py
```diff
@@ -646,6 +475,389 @@ def get_active_tiles(self, assembly, assign=False):
...
+        """ For a tiled Projection, figure out what tiles are hit by an
+        assembly. See parent class documentation for full description.
+
+        """
+        self.compute_nside_tile(assembly) # Set nside_tile if 'auto'
+        return super().get_active_tiles(assembly, assign)
+
+    def get_coords(self, assembly, use_native=False, output=None):
+        """Get the spherical coordinates for the provided pointing Assembly.
+        See parent class documentation for full description.
```
mhasself, line 789: It's not really necessary to redefine this method, right?  There's nothing proj specific about get_coords.  I know the default version uses CAR, but that's true, under the hood, for the HP projector too.

So... please remove this, and just update the inherited get_coords docstring if you feel it is not general enough.

mhasself (approved):
I have a few minor comments before merge, but approving now.

I'm really pleased to have these capabilities in so3g -- thanks again for taking this on!

mhasself:
btw @earosenberg Please squash this on merge (or before).

python/proj/wcs.py
earosenberg: Agree it's not needed in principle. As written there is a technical problem, in that the inherited get_coords will try to initialize a CAR ProjEng using the Healpix _get_pixelizor_args info (nside, etc) and crash. 
I can add a small check to the parent function to call the HP rather than CAR ProjEng, or there are other solutions (eg add a default arg to call 'CAR' with a minimal re-definition in the Healpix case to call 'HP', or a bit more work in the Healpix case to make a proper CAR Projectionist), or else we can leave it as is. Do you have a preference?

mhasself: Ah, right.  Ok, that's a good enough reason.  Leave it in.

mhasself (approved):
Great - squash and rebase and we're done.

Issue #182, opened 2024-06-11
No Python error message for some C++ errors


earosenberg:
Projection routines sometimes fail without a helpful error message.
This seems to have been introduced in commit [acc4ed7](https://github.com/simonsobs/so3g/commit/acc4ed7832d91d7f154d7f309b92896b196d3883).
Two examples are included in this python test file: [test_errors.txt](https://github.com/user-attachments/files/15786891/test_errors.txt)
In the commit prior to the one mentioned the output of these is 
`ValueError: Expected buffer 'signal' to contain items of type float32`
and
`RuntimeError: Buffer 'map' has incompatible shape: Expected (3, *, ->...) but got (1, 250, 300)..`
while in the named commit (and master) it is just `ValueError` and `RuntimeError`.

earosenberg:
Fixed by #183

PR #183, opened 2024-06-14
C++: BufferWrapper fixes (and fix propagation of exception text)

mhasself:
Fixes how array shapes are checked by BufferWrapper (which emerged in #179) ... and tests for that, and additional consequences.

Addresses #182

### DIFF
--- a/include/exceptions.h
+++ b/include/exceptions.h
@@ -15,7 +15,7 @@ public:
     so3g_exception(std::string text) :
         text{text} {}
 
-    std::string msg_for_python() const throw() {
+    virtual std::string msg_for_python() const throw() {
         return text;
     }
 };
--- a/include/numpy_assist.h
+++ b/include/numpy_assist.h
@@ -120,9 +120,7 @@ static std::string shape_string(std::vector<int> shape)
         else if (shape[i] == -1)
             s << "*";
         else if (shape[i] == -2)
-            s << "...->";
-        else if (shape[i] == -3)
-            s << "->...";
+            s << "...";
         else
             s << "!error";
     }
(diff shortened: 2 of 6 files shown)


earosenberg:
This looks good to me -- errors now report properly and the shape checking appears correct. 
I did find two issues when testing -- both appear in the master branch but somewhat related to this PR so I'll mention here:
1) The error "Pointing offset error: nan found at index 1." doesn't appear to be properly "caught" by Python. If it's caught in a try: except: block then subsequent calls to the C++ projection routines (successful or not) give the same error. This is relevant in e.g. the test scripts where for example in test_proj_eng.py `test_00_basic` will fail if the "Raise if pointing invalid" test is done before the others.

2) Healpix to_map segfaults if provided an output map not containing all the pixels needed (ie the array is too short). This can be fixed by passing the expected map size to the shape check routine instead of "-1"
-1 -> npix on lines 510 and 515
-1 -> npix_per_tile on lines 655 and 659
This removes the segfaults (giving a shape error instead) and the other tests still pass, and I don't think we want to allow a non-standard number of pixels in general for healpix.

mhasself:
>     1. The error "Pointing offset error: nan found at index 1." doesn't appear to be properly "caught" by Python. If it's caught in a try: except: block then subsequent calls to the C++ projection routines (successful or not) give the same error. This is relevant in e.g. the test scripts where for example in test_proj_eng.py `test_00_basic` will fail if the "Raise if pointing invalid" test is done before the others.

Hmm... I don't observe that behavior.  Are you sure you reset asm.dets back so it doesn't have any NaNs?   This works fine (on master and on this branch):

```
        # Raise if pointing invalid.                                                                    
        asm.dets[1,2] = np.nan
        with self.assertRaises(ValueError):
           p.to_map(sig, asm, comps='T')

        scan, asm, (shape, wcs) = get_basics()
        p.to_map(sig, asm, comps='T')

        asm.dets[1,2] = np.nan
        with self.assertRaises(ValueError):
           p.to_weights(asm, comps='T')
```

> 
>     2. Healpix to_map segfaults if provided an output map not containing all the pixels needed (ie the array is too short). This can be fixed by passing the expected map size to the shape check routine instead of "-1"
>        -1 -> npix on lines 510 and 515
>        -1 -> npix_per_tile on lines 655 and 659
>        This removes the segfaults (giving a shape error instead) and the other tests still pass, and I don't think we want to allow a non-standard number of pixels in general for healpix.

Ah, good -- I put those in.

Thanks!

earosenberg:
Ah you're right of course, silly mistake. Thanks for the healpix fix too, looks good now I think

earosenberg (approved):

PR #187, opened 2024-10-09
Implement linear interpolation function with GSL

mmccrackan:
This branch adds a simple parallelized linear interpolator using GSL.  Since `FindGSL.cmake` was already in SPT3G, only a few lines needed to be added to the so3g cmake file to include it.  I wrote it to mimic the output of `scipy.interpolate.interp1d` when `kind=linear` and `fill_value='extrapolation'`, which is the setup currently used in the sotodlib azss code.  

GSL has a few other interpolation options (linear, cspline, periodic cspline, non-rounded akima spline) which we can also write functions (or accept an interpolation kind argument maybe) for if we need them.  GSL does not handle extrapolation by itself, so the other options will likely need more careful handling of the extrapolation than what I've done here for now (as highlighted in this discussion: https://docs.scipy.org/doc/scipy/tutorial/interpolate/extrapolation_examples.html#extrapolation-tips-and-tricks).

### DIFF
--- a/src/array_ops.cxx
+++ b/src/array_ops.cxx
@@ -824,11 +824,14 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const 
 }
 
 template <typename T>
-void _gsl_interp(double* x, double* y, double* x_interp, T* y_interp, int n_x, int n_x_interp)
+using _gsl_interp_func_pointer = void (*)(const double*, const double*, const double*, T*, const int, const int,
+                                          gsl_spline*, gsl_interp_accel*);
+
+template <typename T>
+void _gsl_linear_interp(const double* x, const double* y, const double* x_interp, T* y_interp, const int n_x, const int n_x_interp,
+                        gsl_spline* spline, gsl_interp_accel* acc)
 {
-    // these are not thread safe
-    gsl_interp_accel *acc = gsl_interp_accel_alloc();
-    gsl_spline *spline = gsl_spline_alloc(gsl_interp_linear, n_x);
+    // re initialize for each row
     gsl_spline_init(spline, x, y, n_x);
 
     double x_step_left = x[1] - x[0];
@@ -854,13 +857,18 @@ void _gsl_interp(double* x, double* y, double* x_interp, T* y_interp, int n_x, i
             y_interp[si] = gsl_spline_eval(spline, x_interp[si], acc);
         }
     }
-
-    gsl_spline_free(spline);
-    gsl_interp_accel_free(acc);
 }
 
 template <typename T>
-void gsl_linear_interp(const bp::object & x, const bp::object & y, const bp::object & x_interp, bp::object & y_interp)
+void _gsl_cspline_interp(const double* x, const double* y, const double* x_interp, T* y_interp, const int n_x, const int n_x_interp,
+                         gsl_spline* spline, gsl_interp_accel* acc)
+{
+    throw RuntimeError_exception("GSL cspline interpolation is not yet implemented.");
+}
+
+template <typename T>
+void _gsl_interp1d(const bp::object & x, const bp::object & y, const bp::object & x_interp, bp::object & y_interp, 
+                   const gsl_interp_type* interp_type, _gsl_interp_func_pointer<T> interp_func)
 {
     BufferWrapper<T> y_buf  ("y",  y,  false, std::vector<int>{-1, -1});
     const int n_rows = y_buf->shape[0];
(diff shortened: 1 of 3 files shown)


src/array_ops.cxx
```diff
@@ -822,6 +823,168 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
     }
 }
 
+template <typename T>
+using _gsl_interp_func_pointer = void (*)(const double*, const double*, const double*, T*, const int, const int,
```
mhasself, line 827: I think we should drop "gsl" from all the definitions here (this and the functions below).  Anything called "gsl" should come from that library.

src/array_ops.cxx
```diff
@@ -822,6 +823,168 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
...
+                                          gsl_spline*, gsl_interp_accel*);
+
+template <typename T>
+void _gsl_linear_interp(const double* x, const double* y, const double* x_interp, T* y_interp, const int n_x, const int n_x_interp,
+                        gsl_spline* spline, gsl_interp_accel* acc)
```
mhasself, line 832: You're wrapping at 110 chars or so ... please wrap closer to 78 chars.

src/array_ops.cxx
```diff
@@ -911,4 +1074,44 @@ PYBINDINGS("so3g")
             "Args:\n"
             "  flag: flag array (int) with shape (ndet, nsamp)\n"
             "  width: the minimum number of contiguous flagged samples\n");
+    bp::def("gsl_interp1d_linear", gsl_interp1d_linear<float>,
+            "gsl_interp1d_linear(x, y, x_interp, y_interp)"
+            "\n"
+            "Perform linear interpolation over rows of array with GSL."
```
mhasself, line 1080: Add "This function uses OMP to parallelize over the dets axis."

Add "Vector x must be strictly increasing.  Values for x_interp beyond the domain of x will be computed based on extrapolation."

src/array_ops.cxx
```diff
@@ -911,4 +1074,44 @@ PYBINDINGS("so3g")
...
+            "\n"
+            "Args:\n"
+            "  x: independent variable (float32) of data with shape (nsamp,)\n"
+            "  y: data array (float32) with shape (ndet, nsamp)\n"
+            "  x_interp: independent variable for interpolated data (float32) with shape (nsamp_interp,)\n"
+            "  y_interp: interpolated data array (float32) with shape (ndet, nsamp_interp)");
```
mhasself, line 1086: Clarify that y_interp is an output buffer that will be modified/filled by the function.

src/array_ops.cxx
```diff
@@ -911,4 +1074,44 @@ PYBINDINGS("so3g")
...
+            "\n"
+            "Perform linear interpolation over rows of array with GSL."
+            "\n"
+            "Args:\n"
+            "  x: independent variable (float64) of data with shape (nsamp,)\n"
+            "  y: data array (float64) with shape (ndet, nsamp)\n"
+            "  x_interp: independent variable for interpolated data (float64) with shape (nsamp_interp,)\n"
+            "  y_interp: interpolated data array (float64) with shape (ndet, nsamp_interp)");
+    bp::def("gsl_interp1d_cspline", gsl_interp1d_cspline<float>,
+            "gsl_interp1d_cspline(x, y, x_interp, y_interp)"
```
mhasself, line 1098: On second thought we should not include these functions until they are implemented.

If we planned on having a single entry point (`interp(type='spline', ...)`) then doing a partial implementation is more helpful, as it makes a future-proof interface.  But if we're not doing that, better to not pretend there's splines available.

src/array_ops.cxx
```diff
@@ -911,4 +1074,44 @@ PYBINDINGS("so3g")
             "Args:\n"
             "  flag: flag array (int) with shape (ndet, nsamp)\n"
             "  width: the minimum number of contiguous flagged samples\n");
+    bp::def("gsl_interp1d_linear", gsl_interp1d_linear<float>,
```
mhasself, line 1077: Again on the names... I don't think we need to emphasize these come from gsl.  Can you rename the exposed functions "interp1d32" and "interp1d64".  The spline ones can be `cspline1d{32,64}`.

mhasself (changes requested):
Thanks for this!  Nice to see GSL was easy to bring in.

A few comments mostly about names of things... I'm open to discussion on that.

src/array_ops.cxx
mmccrackan: Do you think it might be useful to keep them as "interp1d_linear{32,64}" and "interp1d_cspline{32,64}" instead of just "interp1d{32,64}" and so on as interp1d might refer to a range of interpolation methods?  I think removing gsl from the names is a good idea.

I did try the Boost template overloading to see if we could get rid of the 32 and 64 in the names though I haven't managed to make it work, which I think is due to the fact that both versions have the same boost::python object inputs.

I originally wrote the functions with the string approach you mentioned above (with linear as a default parameter), but I thought separating the functions would be cleaner since we don't need to store and check the allowed types each time the function is called.

mhasself: Yup, sure, go for `interp1d_{method}{width}`.

Instead of Boost overloading you can do something like [this](https://github.com/simonsobs/so3g/blob/b665511745ac71a4176d4fa270e6a8a3ed01e77e/src/array_ops.cxx#L135) to figure out the type of the input array, and call different functions from the boost-wrapped function.  (If you go that route, please fix the pointer abuse in `get_type`... it should `return res`.)

mmccrackan: The suggested typing check works great and I've adjusted the interpolation functions to use it.  I've edited the get_dtype function to return res and decrease the reference count on the actual object ob instead of a.

I think the last push should address all of the other requested changes, but let me know if anything else should be changed.

src/array_ops.cxx
```diff
@@ -822,6 +823,170 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
...
+template <typename T>
+void _interp1d(const bp::object & x, const bp::object & y, const bp::object & x_interp,
+               bp::object & y_interp,  const gsl_interp_type* interp_type,
+               _interp_func_pointer<T> interp_func)
+{
+    BufferWrapper<T> y_buf  ("y",  y,  false, std::vector<int>{-1, -1});
+    const int n_rows = y_buf->shape[0];
+    const int n_x = y_buf->shape[1];
+    if (y_buf->strides[1] != y_buf->itemsize || y_buf->strides[0] != y_buf->itemsize*n_x)
+        throw buffer_exception("y must be C-contiguous along last axis");
```
mhasself, line 875: I recognize this was copied from other code, but it is not right ... buffer_exception expects just the variable name, which it renders as `throw buffer_exception("XYZ") ---> "Argument 'XYZ' does not expose buffer protocol, is not contiguous, or does not export a format."`

You should probably just use `throw ValueError_exception(text)` for these.

The use of "C-contiguous along last axis" is a bit confusing (once again, I see this was taken from elsewhere...).  I think better text would be "Argument 'XYZ' must be a C-contiguous 2d array."

It is possible to relax that constraint and only insist it be contiguous in the 2nd dimension.  That would require making use of strides[0] in the pointer arithmetic.  Very doable -- and helps support partial fills like:
```
   so3g.interp1d_linear(x, y, xx[1000:], yy[:,1000:])
```

mhasself (changes requested):
Almost there -- just a request for better exceptions (mostly).  Strides thing is not high priority but consider it.

src/array_ops.cxx
mmccrackan: Okay, I updated the exceptions and I think it should now properly handle the strides for the input data arrays when slicing.  A new test case for the slicing has been added for this.  I also changed some comments, mostly just to match the upper case of the existing code.

mhasself (approved):
Sorry -- last couple of comments on exception strings...  I'll approve this now though.  Please squash on merge after fixing those error strings.

Thanks for adding the stride handling (and tests!).

src/array_ops.cxx
```diff
@@ -822,6 +823,176 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
...
+}
+
+template <typename T>
+void _interp1d(const bp::object & x, const bp::object & y, const bp::object & x_interp,
+               bp::object & y_interp, const gsl_interp_type* interp_type,
+               _interp_func_pointer<T> interp_func)
+{
+    BufferWrapper<T> y_buf  ("y",  y,  false, std::vector<int>{-1, -1});
+    if (y_buf->strides[1] != y_buf->itemsize)
+        throw ValueError_exception("Argument 'y' must be a C-contiguous 2d array.");
```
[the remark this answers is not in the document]
mmccrackan: Sorry for the delay.  I've added the requested changes.  Could you re-approve as my last push undid that.

PR #188, opened 2024-10-22
Add detrend function

mmccrackan:
This branch adds a detrend function that duplicates the functionality in sotodlib's detrend_tod function for 1D or 2D arrays (sotodlib supports higher dimensions, but I haven't reproduced that here for now).  It has mean, median (using GSL median which is much faster than a std::sort approach), and linear detrending along the rows of an array.

This is primarily intended to be used in an upcoming Welch PSD function but is also standalone, so the underlying function can be run sequentially and called within a parallel loop inside another function (more efficient for the Welch PSD calculation).

### DIFF
--- a/src/array_ops.cxx
+++ b/src/array_ops.cxx
@@ -15,7 +15,10 @@ extern "C" {
 }
 
 #include <boost/python.hpp>
+#include <omp.h>
+
 #include <gsl/gsl_spline.h>
+#include <gsl/gsl_statistics.h>
 
 #include <pybindings.h>
 #include "so3g_numpy.h"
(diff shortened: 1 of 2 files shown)


src/array_ops.cxx
```diff
@@ -1097,4 +1252,17 @@ PYBINDINGS("so3g")
...
+            "Detrend each row of an array (float32/float64). This function uses OMP to parallelize "
+            "over the dets (rows) axis."
+            "\n"
+            "Args:\n"
+            "  tod: input array (float32/float64) buffer with shape (ndet, nsamp) that is to be detrended. "
+            "       The data is modified in place.\n"
+            "  method: how to detrend data.  Options are 'mean', 'median', and 'linear'. Linear calculates "
+            "          and subtracts the slope from either end of each row as determined from 'linear_ncount'.\n"
+            "  linear_ncount: Number (int) of samples to use, on each end, when measuring mean level for 'linear'"
+            "                 detrend. Values larger than 1 suppress the influence of white noise.\n");
```
mhasself, line 1267: Explain that values must be -1 or a positive integer, and -1 causes n_samps / 2 to be used.

mhasself (approved):
Sorry this fell off my radar... looks good.

Approving but please fix the docstring.

Please squash on merge.

src/array_ops.cxx
mmccrackan: Updated.

mhasself (approved):

PR #189, opened 2024-10-29
Add fitting functionality using Ceres-Solver

mmccrackan:
This branch is an attempt at adding fitting functionality to `so3g` using the `Ceres-Solver` library to support both non-linear least squares and general optimization problems.  It is used to duplicate the existing noise-fitting function in `sotodlib`. The idea of this function is to:

1. Find an initial guess for the white noise by calculating the median of the PSD above an input lower frequency bound.
2. Do a least-squares fit of the 1/f component of the PSD to a power law in logspace to estimate the index and knee frequency. 
3. Run the minimization using a negative log likelihood function as the cost function.

The cost function classes and fitting functions have been written in a model-independent manner, so they should be easy to adapt for other data and model types.   `Ceres` includes options for bounds/constraints and supports auto differentiation so derivatives and gradients don't need be calculated manually.  The general minimization routine for `Ceres` doesn't seem to have uncertainty calculations (its least-squares routines do) so I have manually calculated the inverted Hessian matrix to get the parameter uncertainties.

I put most of the code into new `fitting_ops.cxx` and `fitting_ops.h` files instead of putting it all in `array_ops.cxx`.  I added an `array_ops.h` file to hold shared functions declarations.

I've added a cmake directory with cmake files for `Ceres,` `Eigen`, `Gflag`, and `Glog` with the last 3 being dependencies of `Ceres-Solver`.  We could move these into `spt3g` at some point, but this also seems to work in my tests.  `Eigen` is a nice optimized vector, matrix, and linear algebra library that might be useful on its own, but it is not required to use `Eigen` when working with `Ceres`.

Building the Docker image for tests will take a few minutes longer now as I found that Ceres needed to be built from source as the necessary version doesn't appear to be available through `apt-get libceres-dev`.


include/array_ops.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+int get_dtype(const bp::object &);
+
+template <typename T>
+T _calculate_median(const T*, const int);
```
mhasself, line 6: Add newline

include/fitting_ops.h
```diff
@@ -0,0 +1,139 @@
...
+private:
+    const int n_pts;
+    const double* x;
+    const double* y;
+};
+
+// Model independent Negative Log Likelihood for generalized 
+// unconstrained minimization
+template <typename Model>
+struct NegLogLikelihood
```
mhasself, line 103: I think it's important to point out, in the comment and/or name of the class, that this is what you should use for fitting power spectra.  One should use this only on data whose residuals are expected to be Chi2(1)-distributed.  Is it accurate to call it PowerSpectrumCostFunction?

src/fitting_ops.cxx
```diff
@@ -0,0 +1,463 @@
...
+            p[i] = std::numeric_limits<double>::quiet_NaN();
+            if (c) {
+                c[i] = std::numeric_limits<double>::quiet_NaN();
+            }
+        }
+    }
+}
+
+template <typename T>
+bool _invert_matrix(const T* matrix, T* inverse, const int n) {
```
mhasself, line 91: Was this code copyright-free?

src/fitting_ops.cxx
```diff
@@ -0,0 +1,463 @@
...
+             "fit_noise(f, pxx, p, c, lowf, fwhite_lower, fwhite_upper, tol, niter, epsilon)"
+             "\n"
+             "Fits noise model with white and 1/f components to the PSD of signal. Uses a MLE "
+             "method that minimizes a log likelihood. OMP is used to parallelize across dets (rows)."
+             "\n"
+             "Args:\n"
+             "  f: frequency array (float32/64) with dimensions (nsamps,).\n"
+             "  pxx: PSD array (float32/64) with dimensions (ndets, nsamps).\n"
+             "  p: parameter array (float32/64) with dimensions (ndets, nparams). This is modified in place.\n"
+             "  c: uncertaintiy array (float32/64) with dimensions (ndets, nparams). This is modified in place.\n"
```
mhasself, line 452: For p and c -- do the values going in matter?  Or can it just be zeros/empty?

src/fitting_ops.cxx
```diff
@@ -0,0 +1,463 @@
...
+PYBINDINGS("so3g")
+{
+     bp::def("fit_noise", fit_noise,
+             "fit_noise(f, pxx, p, c, lowf, fwhite_lower, fwhite_upper, tol, niter, epsilon)"
+             "\n"
+             "Fits noise model with white and 1/f components to the PSD of signal. Uses a MLE "
+             "method that minimizes a log likelihood. OMP is used to parallelize across dets (rows)."
+             "\n"
+             "Args:\n"
+             "  f: frequency array (float32/64) with dimensions (nsamps,).\n"
```
mhasself, line 449: Are there requirements on f?  Must it be positive definite; must it be strictly increasing?

src/fitting_ops.cxx
```diff
@@ -0,0 +1,463 @@
...
+
+    // index for f < upper fwhite
+    for (int i = nsamps - 1; i >= 0; --i) {
+        if (f[i] < fwhite_upper) {
+            fwhite_i.push_back(i);
+            break;
+        }
+    }
+
+    int fwhite_size = fwhite_i[1] - fwhite_i[0] + 1;
```
mhasself, line 250: This looks segfaulty.  Make it robust against fwhite_lower / fwhite_upper pathologically dodging your tests.  A good way is like
```
int fwhite_lo = 0;
while (fwhite_lo < nsamps && f[fwhite_lo] < fwhite_lower)
  fwhite_lo++;

fwhite_hi = fwhite_lo;
whilte (...
```

mhasself (changes requested):
Very interesting -- thanks for bringing this in and addressing such an important case.

I can't promise these are all my comments, yet, but it's a start...

include/fitting_ops.h
mmccrackan: Apologies for the late response to this.  Since the cost function classes are model-independent, I wanted to keep the names general instead of tying them to the power spectrum model.  I updated the comment to be more descriptive.

src/fitting_ops.cxx
mmccrackan: I didn't copy this just used basic references, so should be good.  I did cross-check the outputs with a function using LAPACK that we could use instead, though that is much slower for these small matrices.

src/fitting_ops.cxx
mmccrackan: Updated comment to explain that the input values are ignored.

src/fitting_ops.cxx
mmccrackan: I think it would have failed previously due to the segfault-prone code you pointed out, but should be fine now.  I added some description to the docstring to be more informative.

src/fitting_ops.cxx
mmccrackan: Updated.

include/array_ops.h
mmccrackan: Done.

src/fitting_ops.cxx
```diff
@@ -0,0 +1,469 @@
...
+
+    // Index where f = fwhite_upper
+    fwhite_i[1] = nsamps - 1;
+    while (fwhite_i[1] > fwhite_i[0] && f[fwhite_i[1]] > fwhite_upper) {
+        fwhite_i[1]--;
+    }
+
+    if (fwhite_i[0] < lowf) {
+        throw ValueError_exception("fwhite lower index < lower freq index.");
+    }
```
mhasself, line 247: I believe this compares an index to a frequency.

When I see a function like this I'm inclined to rewrite it generically -- consider a function that takes double *f and also a vector of f_lims, both assumed increasing, and returns vector of f_lims_idx, with the indices of those f_lims.  There's a lot of specificity here that isn't really needed.  And it would clarify that in all cases you are getting the lowest index `i` s.t. `f[i] >= f_lim`  (not even sure if that's the case here -- because each search loop is written somewhat differently).

src/fitting_ops.cxx
```diff
@@ -0,0 +1,469 @@
...
+             "Fits noise model with white and 1/f components to the PSD of signal. Uses a MLE "
+             "method that minimizes a log-likelihood. OMP is used to parallelize across dets (rows)."
+             "\n"
+             "Args:\n"
+             "  f: frequency array (float32/64) with dimensions (nsamps,). "
+             "     Should be positive definite and strictly increasing.\n"
+             "  pxx: PSD array (float32/64) with dimensions (ndets, nsamps).\n"
+             "  p: output parameter array (float32/64) with dimensions (ndets, nparams). "
+             "     This is modified in place and input values are ignored.\n"
+             "  c: output uncertaintiy array (float32/64) with dimensions (ndets, nparams). "
```
mhasself, line 457: typo

mhasself (commented):
Thanks!  Couple minor comments.

Also, can you add a test where the data are noisy, i.e. where the data have Chi2-1 dist from the model?  You can do that by just drawing from normal(model(f)**.5) and then squaring the results.

src/fitting_ops.cxx
mmccrackan: Yeah that was a typo.  I've generalized the function as suggested and do up-front checks on the input frequency values.

src/fitting_ops.cxx
mmccrackan: Fixed.

mmccrackan:
> Thanks! Couple minor comments.
> 
> Also, can you add a test where the data are noisy, i.e. where the data have Chi2-1 dist from the model? You can do that by just drawing from normal(model(f)**.5) and then squaring the results.

Added the test with noise. It was a good thing you suggested this as I found that the logger needs to be shut down at the end otherwise it will fail if the fit function is called twice.

mhasself (approved):

Issue #190, opened 2024-10-31
Tiled Healpix FloatingPointError for bad nside_tile


earosenberg:
A FloatingPointError is raised in Projection.cxx when trying to run to_map (and probably other operations) with a Pixelizor_Healpix<Tiled> with nside_tile > nside.

This can happen when nside_tile='auto' is used in the Python layer if nside is small and OMP_NUM_THREADS is large.

PR #191, opened 2024-11-01
Fix healpix 'auto' tiling

earosenberg:
Addresses  #190.  
The 'auto' routine for setting the 'Healpix tiling now caps nside_tile to nside.
Add check that nside_tile <= nside in C++. If not it will raise a ValueError that can be passed back to Python instead of failing on a FloatingPointError.

### DIFF
--- a/src/Projection.cxx
+++ b/src/Projection.cxx
@@ -596,6 +596,11 @@ public:
         npix_per_tile = npix / ntiles;
         check_nside(nside); // check validity
         check_nside(nside_tile);
+        if (nside_tile > nside) {
+            std::ostringstream err;
+            err << "Invalid nside_tile " << nside_tile << " > nside " << nside;
+            throw ValueError_exception(err.str());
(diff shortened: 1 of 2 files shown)


src/Projection.cxx
```diff
@@ -596,6 +596,11 @@ class Pixelizor_Healpix<Tiled> {
         npix_per_tile = npix / ntiles;
         check_nside(nside); // check validity
         check_nside(nside_tile);
+        if (nside_tile > nside){
+          std::ostringstream err;
+          err << "Invalid nside_tile " << nside_tile << " > nside " << nside;
+          throw ValueError_exception(err.str());
+        }
```
mhasself, line 603: Thanks, just fix the indentation (4-space), and add a space before the `{`.

earosenberg: Fixed, thanks

mhasself (approved):

PR #192, opened 2024-11-21
Update build system to support latest upstream spt3g

tskisner:
- Switch to std::shared_ptr

- Attempt to build internal libso3g with _libso3g wrapper extension

PR #193, opened 2024-12-01
Stokes response

amaurea:
Edit: No longer breaks backwards compatibility - see jan 10 comment.

The pointing matrix in so3g has had a hardcoded polarization efficiency of 100%. This pull request updates it to support arbitrary T and P responses per detector. This was needed to be able to represent demodulated timestreams as virtual detectors, which will be a separate pull request for sotodlib later. The virtual detector approach to demodulation mapmaking has the advantage that none of the rest of the code needs to know about the demodulation, only the demodulation function itself.

This breaks backwards compatibility in the interface of the FocalPlane object. It used to be an OrderedDict of `[name, q]` where q is an entry in a `G3VectorQuat`, but it is now a class with the members `.quats`, which is an [ndet,4] array of quaternion coefficients, and `.resps`, which is an `[ndet,2]` array of T and P responses. This means that FocalPlane no longer supports named detectors. This simplifies `Assembly`, which now just contains a FocalPlane instance.

This incompatibility will require corresponding changes in sotodlib. Do we have to go through a slow dance of add compatibility layer in sotodlib, wait for a few weeks, update so3g, then remove compatibility layer in sotodlib, again? I hope not. What do you think?

### DIFF
--- a/docs/proj.rst
+++ b/docs/proj.rst
@@ -85,9 +85,9 @@ quaternions into rotation angles::
   >>> csl.Q
   spt3g.core.G3VectorQuat([(-0.0384748,0.941783,0.114177,0.313891)])
 
-  >>> csl.coords()   
+  >>> csl.coords()
   array([[ 0.24261138, -0.92726871, -0.99999913, -0.00131952]])
-  
+
 The :func:`coords() <CelestialSightLine.coords>` returns an array with
 shape (n_time, 4); each 4-tuple contains values ``(lon, lat,
 cos(gamma), sin(gamma))``.  The ``lon`` and ``lat`` are the celestial
@@ -117,12 +117,12 @@ orientations::
 
 This particular function, :func:`from_xieta()
 <FocalPlane.from_xieta>`, will apply the SO standard coordinate
-definitions and stores quaternion coefficients (``.quats``) and
-responsivities (``.resps``) for each detectors. These are numpy
-arrays with shape ``[ndet,4]`` and ``[ndet,2]`` respectively.
-``fp.quats[0]`` gives the 4 quaternion coefficients for the
-first detector, while ``fp.resps[0]`` gives its total intensity
-and polarization responsivity.::
+definitions and stores quaternions (``.quats``) and
+responsivities (``.resps``) for each detector. These are a
+``G3VectorQuat``  wth length ``ndet`` and a numpy array
+with shape ``[ndet,2]`` respectively. ``fp.quats[0]`` gives
+the quaternion for the first detector, while ``fp.resps[0]``
+gives its total intensity and polarization responsivity.::
 
   >>> fp.quats[2]
   array([  0.86601716,  0.00377878, -0.00218168,  0.49999524])
@@ -133,7 +133,7 @@ As you can see, the default responsivity is 1 for both total
 intensty and polarization. Note that ``.quats`` contains
 quaternion *coefficients*, not quaternion *objects*. To do
 quaternion math, you need to convert them to actual quaternion
-objects, e.g. ``q = spt3g.core.quat(*fp.quats[0])``, or convert
+objects, e.g. ``q = fp.quats[0]``, or convert
 them all at once with ``qs = spt3g.core.G3VectorQuat(fp.quats)```.
 
 To represent detectors with responsivity different from 1,
@@ -166,7 +166,7 @@ At this point you could get the celestial coordinates for any one of
 those detectors::
 
   # Get vector of quaternion pointings for detector 0
-  q_total = csl.Q * spt3g.core.quat(*fp.quats[0])
+  q_total = csl.Q * fp.quats[0]
   # Decompose q_total into lon, lat, roll angles
   ra, dec, gamma = so3g.proj.quat.decompose_lonlat(q_total)
 
@@ -185,9 +185,8 @@ FocalPlane object as first argument::
    array([[ 0.24261138, -0.92726871,  0.86536489,  0.5011423 ]]),
    array([[ 0.22806774, -0.92722945,  0.50890634,  0.86082189]])]
 
-To be clear, ``coords()`` now returns a dictionary whose keys are the
-detector names.  Each value is an array with shape (n_time,4), and at
-each time step the 4 elements of the array are: ``(lon, lat,
+So ``coords()`` returns a list of numpy arrays with shape (n_time,4),
+and at each time step the 4 elements of the array are: ``(lon, lat,
 cos(gamma), sin(gamma))``.
 
 
--- a/docs/proj.rst
+++ b/docs/proj.rst
@@ -110,26 +110,63 @@ Pointing for many detectors
 Create a :class:`FocalPlane` object, with some detector positions and
 orientations::
 
-  names = ['a', 'b', 'c']
-  x = np.array([-0.5, 0., 0.5]) * DEG
-  y = np.zeros(3)
+  xi  = np.array([-0.5, 0.0, 0.5]) * DEG
+  eta = np.zeros(3)
   gamma = np.array([0,30,60]) * DEG
-  fp = so3g.proj.FocalPlane.from_xieta(names, x, y, gamma)
+  fp  = so3g.proj.FocalPlane.from_xieta(xi, eta, gamma)
 
 This particular function, :func:`from_xieta()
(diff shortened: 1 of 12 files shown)


amaurea:
Looks like some other code needs to be updated to support the new FocalPlane class that replaced the old OrderedDict. Some places expect to be able to iterate over `.items()` and get names and `G3VectorQuat`s out. How could this best be handled? The most quick-and-dirty solution would be to implement `.items()` and out it spit out things like that. However, this interface can't capture the new response stuff, so I fear that this would hide code that actually should fail because it needs to be updated to take the responses into account.

amaurea:
This pull request is now backwards compatible with the current version of sotodlib. Let's get it merged so I can get my new sotodlib branch merged! It's needed for my demod_ml_mapmaker for example.

tskisner:
I'm not as familiar with this part of so3g, so perhaps better to have @mhasself review.  I'll just note that there may be some conflicts with the new relative imports in the build system PR, so I will plan to rebase #192 after this PR is merged.

python/proj/coords.py
```diff
@@ -208,93 +208,205 @@ def for_horizon(cls, t, az, el, roll=None, site=None, weather=None):
...
         Returns:
-          If the det_offset was passed in as a dict, then a dict with the same
-          keys is returned.  Otherwise a list is returned.  For each
-          detector, the corresponding returned object is an array with
-          shape (n_samp, 4) where the four components correspond to
-          longitude, latitude, cos(psi), sin(psi).
-
+          If fplane is not None, then the result will be
+          [n_samp,{lon,lat,cos2psi,sin2psi}]. Otherwise it will
+          be [n_det,n_Samp,{lon,lat,cos2psi,sin2psi}]
```
mhasself, line 224: Remove "not".

n_Samp -> n_samp

python/proj/wcs.py
```diff
@@ -460,7 +460,7 @@ def get_active_tiles(self, assembly, assign=False):
             #     print(f"Warning: Threads poorly balanced. Max/mean hits = {max_ratio}")
 
             # Now paint them into Ranges.
-            R = projeng.tile_ranges(q_native, assembly.dets, group_tiles)
+            R = projeng.tile_ranges(q_native, assembly.fplane.quats, group_tiles)
```
mhasself, line 463: You missed the last one ... line 798.  Also the comment on line 29.

test/test_proj_eng_hp.py
```diff
@@ -23,7 +23,7 @@ def get_scan():
 def get_basics():
     t, az, el = get_scan()
     csl = proj.CelestialSightLine.az_el(t, az, el, weather='vacuum', site='so')
-    fp = proj.FocalPlane.from_xieta(['a', 'b'], [0., .1*DEG], [0, .1*DEG])
+    fp = proj.FocalPlane.from_xieta([0., .1*DEG], [0, .1*DEG])
```
mhasself, line 26: Also please fix demos/demo_proj_hp.py ... this does it:
```
--- a/demos/demo_proj_hp.py
+++ b/demos/demo_proj_hp.py
@@ -191,13 +191,6 @@ def simulate_detectors(ndets, rmax, seed=0):
     eta = rr * np.sin(phi)
     return xi, eta, gamma
 
-def make_fp(xi, eta, gamma):
-    fp = so3g.proj.quat.rotation_xieta(xi, eta, gamma)
-    so3g_fp = so3g.proj.FocalPlane()
-    for i, q in enumerate(fp):
-        so3g_fp[f'a{i}'] = q
-    return so3g_fp
-
 def extract_coord(sight, fp, isignal=0, groupsize=100):
     coord_out = []
     for ii in range(0, len(fp), groupsize):
@@ -207,15 +200,14 @@ def extract_coord(sight, fp, isignal=0, groupsize=100):
     return coord_out
 
 def make_signal(ndets, rmax, sight, isignal, seed=0):
-    xi, eta, gamma = simulate_detectors(ndets, rmax, seed)
-    fp = so3g.proj.quat.rotation_xieta(xi, eta, gamma)
+    fp = so3g.proj.FocalPlane.from_xieta(*simulate_detectors(ndets, rmax, seed))
     signal = extract_coord(sight, fp, isignal)
     return signal
 
 def make_tod(time, az, el, ndets, rmax, isignal, seed=0):
     sight = so3g.proj.CelestialSightLine.naive_az_el(time, az, el)
     signal = make_signal(ndets, rmax, sight, isignal, seed)
-    so3g_fp = make_fp(*simulate_detectors(ndets, rmax, seed))
+    so3g_fp = so3g.proj.FocalPlane.from_xieta(*simulate_detectors(ndets, rmax, seed))
     asm = so3g.proj.Assembly.attach(sight, so3g_fp)
     return signal, asm
```

mhasself (commented):
Thanks for this, and for the iterations (offline) on the docs.  Couple minor things inline... the only larger concern is whether focalplane.quats should be an array, or ... quats.

docs/proj.rst
```diff
@@ -110,44 +110,80 @@ Pointing for many detectors
...
 
 At this point you could get the celestial coordinates for any one of
 those detectors::
 
-  # Get vector of quaternion pointings for detector 'a'
-  q_total = csl.Q * fp['a']
+  # Get vector of quaternion pointings for detector 0
+  q_total = csl.Q * spt3g.core.quat(*fp.quats[0])
   # Decompose q_total into lon, lat, roll angles
   ra, dec, gamma = so3g.proj.quat.decompose_lonlat(q_total)
```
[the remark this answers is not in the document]
amaurea: I thought about the same thing when preparing the pull request. I think the reason was that the C++ code needs a numpy array. Ideally G3VectorQuat would have the same memory layout as the corresponding numpy array of coefficients, so conversion would be free, but as it is, one has to do a conversion any time one wants to use any pointing operation. I wanted the internal representation to have no impedance mismatch with the low-level pointing code, and that's why I ended up with the numpy array.

That said, this is a small array. The conversion cost will be completely negligible unless we have very few samples. Let's just switch to it being a G3VectorQuat and eating the conversion cost.

docs/proj.rst
```diff
@@ -110,44 +110,80 @@ Pointing for many detectors
...
 FocalPlane object as first argument::
 
   >>> csl.coords(fp)
-  OrderedDict([('a', array([[ 0.22806774, -0.92722945, -0.9999468 ,
-  0.01031487]])), ('b', array([[ 0.24261138, -0.92726871, -0.86536489,
-  -0.5011423 ]])), ('c', array([[ 0.25715457, -0.92720643, -0.48874018,
-  -0.87242939]]))])
+  [array([[ 0.25715457, -0.92720643,  0.9999161 ,  0.01295328]]),
+   array([[ 0.24261138, -0.92726871,  0.86536489,  0.5011423 ]]),
+   array([[ 0.22806774, -0.92722945,  0.50890634,  0.86082189]])]
```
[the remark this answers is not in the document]
amaurea: Fixing. We should check whether its sin(gamma) or sin(2*gamma) at some point too. I'm leaving in the sin(gamma) for now, since I don't know that it's wrong.

python/proj/coords.py
```diff
@@ -208,93 +208,205 @@ def for_horizon(cls, t, az, el, roll=None, site=None, weather=None):
...
         """
-        qs = quat.rotation_xieta(np.asarray(xi), np.asarray(eta), np.asarray(gamma))
-        return cls([(n,q) for n, q in zip(names, qs)])
-
+        # The underlying code wants polangle gamma and the T and P
+        # response, but we support speifying these as the T, Q and U
+        # response too. Writing it like this handles both cases, and
+        # as a bonus(?) any combination of them
+        # FIXME: old sotodlib compat - remove later
+        xi, eta, gamma, T, P, Q, U, hwp, dets = cls._xieta_compat(xi,eta,gamma,T,P,Q,U,hwp)
```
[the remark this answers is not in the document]
amaurea: I fixed it in a more direct way by manipulating *args and **kwargs directly, basically just removing the first element of args.

python/proj/coords.py
amaurea: Fixed

mhasself (dismissed):
Last couple of minor things...  but approving.

Squash this on merge!!

mhasself (approved):
Hooray!  So stoked!

Squash on merge!

PR #195, opened 2024-12-17
Slight fix for GSL interpolation parallelization

mmccrackan:
After trying out the GSL interpolation in `sotodlib` with the recent release of `so3g`, I noticed that the performance was not scaling with thread count when the GSL initialization is done per-thread.  After some testing, it seems that doing the initialization in the parallelized for loop results in the best performance scaling with threads.


mhasself (approved):
If that latest correction fixes the performance issue, you have my blessing.

Squash on merge!

mmccrackan:
> If that latest correction fixes the performance issue, you have my blessing.
> 
> Squash on merge!

Tested it and it works great!  Thanks for pointing out the mistake!

PR #196, opened 2024-12-26
Add c++ jump fixing and finding of slow jumps

skhrg:
Still a draft because my so3g build system is broken rn...

### DIFF
--- a/src/array_ops.cxx
+++ b/src/array_ops.cxx
@@ -6,6 +6,7 @@
 #include <vector>
 #include <string>
 #include <cmath>
+#include <algorithm>
 extern "C" {
     #include <cblas.h>
     // Additional prototypes for Fortran LAPACK routines.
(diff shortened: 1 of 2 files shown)


src/array_ops.cxx
```diff
@@ -823,6 +882,52 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
...
+    #pragma omp parallel for
+    for(int di = 0; di < ranges.size(); di++) {
+        int start = 0;
+        int stop = 0;
+        T min_h;
+        T max_h;
+        T height;
+        T to_sub = 0;
+        for (auto const &r: ranges[di].segments) {
+            start = di*nsamp + r.first;
```
kwolz, line 912: Is it ensured that every r.first (lower end of the jump range) is greater than r.first of the preceding one? I think this is needed for this code to work.

kwolz (commented):
Looks great! Only a single comment.

src/array_ops.cxx
skhrg: Yes, I believe that this is guaranteed by the ranges matrix. Ranges cannot overlap so the `r.first` of a range will always be at least 1 greater than the `r.second` of the preceding one.

src/array_ops.cxx
```diff
@@ -1074,6 +1196,25 @@ PYBINDINGS("so3g")
             "Output will be 0 where jumps are not found and the assumed jump height where jumps are found.\n"
             "\n"
             "See details in docstring for find_quantized_jumps.\n");
+    bp::def("subtract_jump_heights", subtract_jump_heights<float>,
+            "subtract_jump_heights(tod, out, heights, jumps)"
+            "\n"
+            "Subtract fit jump heights from known jump locatations in a float32 array."
```
mhasself, line 1202: Typo "locations".

The description here doesn't really tell someone how to use this function.  Consider expanding this docstring to actually say what is done.  At the very least you should make it more clear that the jumps are treated _cumulatively_, potentially modifying almost every sample.  Something like:

"For each detector, compute the cumulative effect of the jumps identified by the array "heights" and the RangesMatrix "jumps".  For each range in "jumps", the values from "heights" are checked and the size of the jump is either the largest positive or the largest negative number (whichever has the largest absolute value).  The "output" value is the difference of "tod" and the accumulated jump vector."

mhasself (changes requested):
Very happy to see this!

Please add smoke tests for these functions.

I would really like you to loosen the requirement on strides[0].  But I suppose it's not urgent for me if it's not urgent for you.

src/array_ops.cxx
```diff
@@ -823,6 +882,52 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
...
+        T to_sub = 0;
+        for (auto const &r: ranges[di].segments) {
+            start = di*nsamp + r.first;
+            for(int j = stop; j < start && to_sub != 0; j++) {
+                output[j] = tod_data[j] - to_sub;
+            }
+            stop = di*nsamp + r.second;
+            min_h = *(std::min_element(jump_heights+start, jump_heights+stop));
+            max_h = *(std::max_element(jump_heights+start, jump_heights+stop));
+            height = (max(abs(min_h), abs(max_h)) == abs(min_h)) ? min_h : max_h;
```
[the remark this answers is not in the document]
skhrg: Yeah does that test and also sets the value in the same line (also guards against the case where one is an overflow I think)

src/array_ops.cxx
skhrg: @mhasself can you confirm that what I said above is true?

src/array_ops.cxx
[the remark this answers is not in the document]
mhasself: Exponential question growth!!  What is "the case where one is an overflow"?  In what sense does this guard against it?  And, are we sure that "guarding against it" is what we want to do, in that case?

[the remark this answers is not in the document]
skhrg: Actually realizing I cut the overflow stuff, that would be:
 `height = (max_h == FLT_MIN ||  min_h == FLT_MIN) ? FLT_MIN : ((max(abs(min_h), abs(max_h)) == abs(min_h)) ? min_h : max_h);`

The reason for including this is because `abs(FLT_MIN)` overflows (I'm pretty sure)  once you flip the sign, I cut it because:

1. If we are in a state to trigger that it means something is going wrong and we probably want to crash anyways
2. I was too lazy to figure out how to generalize since this is a templated function (ie: for doubles you want `DBL_MIN` instead).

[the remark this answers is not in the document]
mhasself: Ok, thanks for making me learn about limits and overflows...

The C++ templated version of FLT_MAX is `std::numeric_limits<T>::max()`, from `#include <limits>`.

Note that FLT_MIN is not negative-almost-infinity.  Rather it's the smallest representable positive number.  So that's not something we'd worry about.

Yes, leaving it simple should just cause things to quietly saturate at +inf / -inf.  That's fine.

Since you have brought me down this rabbit hole, please change the line to:
```
// Decide whether this is a negative or positive jump.
height = (abs(min_h) > abs(max_h)) ? min_h : max_h;
```

That is the payment I demand.

src/array_ops.cxx
```diff
@@ -823,6 +882,53 @@ void find_quantized_jumps(const bp::object & tod, const bp::object & out, const
...
+        for (auto const &r: ranges[di].segments) {
+            start = di*nsamp + r.first;
+            for(int j = stop; j < start && to_sub != 0; j++) {
+                output[j] = tod_data[j] - to_sub;
+            }
+            stop = di*nsamp + r.second;
+            min_h = *(std::min_element(jump_heights+start, jump_heights+stop));
+            max_h = *(std::max_element(jump_heights+start, jump_heights+stop));
+            // Decide whether this is a negative or positive jump.
+            height = (abs(min_h) > abs(max_h)) ? min_h : max_h;
```
mmccrackan, line 920: I think the answer is yes, but is this okay if there is a jump at the first or last samples?

mmccrackan (approved):
I don't think I have any major comments outside of one quick question.

src/array_ops.cxx
skhrg: Yeah thats fine, the mean level will shift to around the first sample's value in that case, but that should be ok since we will want to mean sub after gapfilling anyways.

mhasself (approved):
Thanks!

I am only looking the other way on the bulk-reformatting of that test file because it is well-separated from where you made actual changes.

Remember to squash on/before merge.

skhrg:
> I am only looking the other way on the bulk-reformatting of that test file because it is well-separated from where you made actual changes.

Stupid diffs giving away my love for black...

PR #201, opened 2025-01-30
Prune unused

mhasself:
Removes a bunch of stuff -- especially compiled stuff -- that is not used:

- G3Ndarray, G3Ndmap, G3WCS; and configurable monkey patching / config stuff in so_frame.py, config.py
- Ranges and Intervals classes are no longer G3Frame objects.
- G3IndexedReader is a shimmed version of G3Reader, which supports seek/tell.  Use of G3IndexedReader is deprecated.
- Rebundler and Reframer -- two reframing codes that are not used and not likely to be.
- quicksim.py -- a kind of cool data simulation generator that we really have no use for
 
This leaves the only G3FrameObjects in so3g as:
- G3SuperTimestream - in active use in SO  data 
- IrregBlockDouble - no longer being produced, but exists in HK data files produced in SO labs up to late 2020.


tskisner (approved):
Thanks for this work in removing unused G3 compiled code and switching to upstream classes.  I verified that the wheel builds work with this branch.

PR #206, opened 2025-02-24
docs: docstring for G3SuperTimestream.Options

### LINKED ISSUE #142: Compression Enable is overrided by data_algo
The order in which compression enable / data algorithms seems a bit strange to me. I would expect that the following line would result in no compression being applied
```
data.Options(enable=0, ..., data_algo=3)
```
however it seems that the [way things are written now](https://github.com/simonsobs/so3g/blob/909d758ce96a09c7b9f51f542a98dd5627ce276d/src/G3SuperTimestream.cxx#L743-L745), the `data_algo` not being 0 overrides the enable option.

If this is the intended behavior feel free to just close this issue.

mhasself:
Resolves #142

### DIFF
--- a/src/G3SuperTimestream.cxx
+++ b/src/G3SuperTimestream.cxx
@@ -814,6 +814,12 @@ bool G3SuperTimestream::Extract(
 	return true;
 }
 
+/* Options()
+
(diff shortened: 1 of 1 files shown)


tskisner (approved):
Looks good.

PR #211, opened 2025-04-25
Transition to nanobind and remove compiled dependency on spt3g

tskisner:
Now that G3SuperTimestream is load-able from upstream spt3g, migrate so3g to a simpler structure without depending on the compiled spt3g package or boost.


tskisner:
After offline discussion, @mhasself would prefer a minimally invasive port from boost::python to nanobind.  Closing this now in favor of a different approach which involves only changes that are absolutely necessary.

PR #213, opened 2025-06-03
perf: add option to skip check that all items have same shape and skip the check in _gibasic

skhrg:


mhasself (approved):
Thanks for finding this!

PR #214, opened 2025-06-20
Update wheel dependencies and manylinux build container

tskisner:


tskisner:
@mhasself or @mmccrackan , if one of you has time to review this it would be much appreciated.  You can see the wheel build tests passing [in this manually-triggered workflow](https://github.com/simonsobs/so3g/actions/runs/16038363503).  After this is merged we can do a tag.  Unless we are trying to get #209 and #210 merged before the next tag?

tskisner:
Actually nevermind, we should do a tag after this is merged before trying to merge those other PRs.

mhasself (approved):
That's great.  If you have no concerns about the wheel actually working, then neither do I!  Thanks for figuring this out.

PR #240, opened 2026-04-17
Fix incorrect implementation of bilinear interpolation

amaurea:
Both the pixel index calculation and the weights calculation were wrong. I must have been asleep when implementing this originally.

### DIFF
--- a/src/Projection.cxx
+++ b/src/Projection.cxx
@@ -51,6 +51,8 @@ inline bool isNone(const bp::object &pyo)
     return (pyo.ptr() == Py_None);
 }
 
+int ifloor(double x) { return int(x)-int(x<0); }
+int iround(double x) { return ifloor(x+0.5); }
 
 // ProjEng template system
 //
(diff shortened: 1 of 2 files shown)


skhrg (dismissed):
I am going to assume the .5 pix difference is correct, looks like a normal corner vs center of pixel thing.

The order of the weights in the interpolation now match what I have written in other places (checked against my thesis draft).  So that now looks good as well.

Maybe more of a @mhasself question but it may be worth it to write a simple unit test to protect against regressions here?

mhasself (changes requested):
Looks good to me, thanks.

@amaurea has agreed to make a quick unit test for this, then we'll get it merged and tagged.

amaurea:
@skhrg Yes, the 0.5 thing is because for nearest neighbor you want to round to the closest integer, while for bilinear you want to find the pixel just before (floor) and just after (ceil = floor+1). I've done this right before and after, so I don't know what I was thinking that time.

mhasself (approved):
Thanks!

PR #241, opened 2026-04-22
Speed up RangesMatrix

mhasself:
Addresses a few slow things ...
- len was implemented as .shape[0], which is not good.
- getitem[int] now avoids all the slicing logic.
- .ones / .zeros / .full now skips shape checking.
- C++ code dealing with vector<Ranges> caches len and preallocates vectors.

With very rough profiling, these changes speed up processing of complex RangesMatrix objects, in C++ code, by a factor of 30 or so.

Thanks to @skhrg for identifying the inefficiencies here!


skhrg (approved):
Looks good to me

PR #245, opened 2026-07-03
fix: use 64 bit indices to avoid overflows

### LINKED ISSUE #244: 32-bit integer overflow of so3g.block_minmax
This leads to Segmentation fault (core dumped).
```
import so3g
import numpy as np
a = np.zeros((4050, 603260), dtype='float32')
so3g.block_minmax(a, a, 800, 2, 0)
```
This is used for slow_jumps of sotodlib
https://github.com/simonsobs/sotodlib/blob/a6176f052d392019b50dc85b994ba47cf5098058/sotodlib/tod_ops/jumps.py#L601
When we analyze single wafer, the number of dets x samps is small enough and we do not have segmentation fault, but we have issue when we analyze multiple wafers at the same time. 

I seem to have same issue for so3g.subtract_jump_heights etc
https://github.com/simonsobs/sotodlib/blob/a6176f052d392019b50dc85b994ba47cf5098058/sotodlib/tod_ops/jumps.py#L211

skhrg:
Fixer #244

### DIFF
--- a/src/array_ops.cxx
+++ b/src/array_ops.cxx
@@ -1,3 +1,4 @@
+#include <cstdint>
 #define NO_IMPORT_ARRAY
 
 #include <complex>
@@ -579,9 +580,9 @@ void test_buffer_wrapper(const bp::object array,
 
 
 template <typename T>
-void _moment(T* data, T* output, int moment, bool central, int start, int stop)
+void _moment(T* data, T* output, int moment, bool central, int64_t start, int64_t stop)
 {
-    int bsize = stop - start;
+    int64_t bsize = stop - start;
     // Could replace the loops with boost accumulators?
     T center = 0.0;
     if(central || moment == 1) {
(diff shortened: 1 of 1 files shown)


src/array_ops.cxx
```diff
@@ -609,19 +610,19 @@ void _moment(T* data, T* output, int moment, bool central, int start, int stop)
 template <typename T>
 void _block_moment(T* tod_data, T* output, int bsize, int moment, bool central, int ndet, int nsamp, int shift)
 {
-    int nblock = ((nsamp - shift)/bsize) + 1;
+    int64_t nblock = (int64_t)(nsamp - shift + bsize - 1) / (int64_t)bsize;
```
ykyohei: Is -1 unnecessary?

src/array_ops.cxx
```diff
@@ -667,19 +668,19 @@ void _minmax(T* data, T* output, int mode, int start, int stop)
 template <typename T>
 void _block_minmax(T* tod_data, T* output, int bsize, int mode, int ndet, int nsamp, int shift)
 {
-    int nblock = ((nsamp - shift)/bsize) + 1;
+    int64_t nblock = (int64_t)(nsamp - shift + bsize - 1) / (int64_t)bsize;
```
ykyohei: same as above

ykyohei (commented):
Thank you for fixing this!

mhasself (dismissed):
Isn't this the purpose of size_t?  It's more self-documenting to use that.  I won't insist now though...

mhasself:
> Isn't this the purpose of size_t? It's more self-documenting to use that. I won't insist now though...

Whoops -- but please address @ykyohei 's question about the nblock calculations.  If this change is truly needed, there should be a unit test to prove it.

skhrg:
> Isn't this the purpose of size_t? It's more self-documenting to use that. I won't insist now though...

I had a vague recollection that `size_t` was platform specific and once upon a time caused a bug for me when [Mist](https://docs.scinet.utoronto.ca/index.php/Mist) was still around. But I guess we will probably never run on a Power CPU ever again...

src/array_ops.cxx
skhrg: Looking at it more carefully I am going to revert this for some reason when re-reading this code I thought I had a `<=` on line 623...

skhrg:
@mhasself @ykyohei any opposition to merging?

ykyohei (approved):
No, looks good to me, thank you!

skhrg:
Oh wait I don't have permission.. I guess I always had Matthew merge for me...

ykyohei:
@mhasself Would it be possible to release a new version of so3g? I would really appreciate it if we could apply this fix on PyPI.

mhasself:
> @mhasself Would it be possible to release a new version of so3g? I would really appreciate it if we could apply this fix on PyPI.

It's on its way ...

src/array_ops.cxx
mhasself, line 662: I think this is the one you missed!

PR #247, opened 2026-07-09
Make code safe for large (>2G) buffers

skhrg:


include/numpy_assist.h
```diff
@@ -108,7 +109,8 @@ inline int numpysafe_extract_int(const bp::object obj, const std::string argstr)
 }
 
 
-static std::string shape_string(std::vector<int> shape)
+template <typename T>
+static std::string shape_string(std::vector<T> shape)
```
mhasself, line 113: Any reason not to just change this to `<int64_t>`?

include/numpy_assist.h
mhasself, line 171: Constructor can/should expect a vector<int64_t> shape, now, right?

src/array_ops.cxx
```diff
@@ -81,7 +81,7 @@ void nmat_detvecs_apply(const bp::object & ft, const bp::object & bins, const bp
         //#pragma omp parallel for
         for(int di = 0; di < ndet; di++)
             for(int i = b1; i < b2; i++)
-                ft_[di*nmode+i] *= biD[di]/norm;
+                ft_[(int64_t)(di*nmode+i)] *= biD[di]/norm;
```
mhasself: Not sufficient ... must cast one of the multiplicands directly; e.g.
`ft_[(int64_t)di*n_mode+i] ...`

Note also that `(*ft_buf.ptr_2d(di, i))` would get this same reference.

src/array_ops.cxx
```diff
@@ -250,22 +250,22 @@ void pcut_full_tod2vals_helper(const vector<RangesInt32> & rangemat, T * tod, in
     for(int di = 0; di < rangemat.size(); di++)
         for (auto const &r: rangemat[di].segments)
             for(int j = r.first; j < r.second; j++, i++)
-                vals[i] = tod[di*nsamp+j];
+                vals[i] = tod[(int64_t)(di*nsamp+j)];
```
mhasself: Same -- apply the cast to di.

src/array_ops.cxx
```diff
@@ -250,22 +250,22 @@ void pcut_full_tod2vals_helper(const vector<RangesInt32> & rangemat, T * tod, in
...
 }
 template <typename T>
 void pcut_full_vals2tod_helper(const vector<RangesInt32> & rangemat, T * tod, int nsamp, T * vals) {
     int i = 0;
     for(int di = 0; di < rangemat.size(); di++)
         for (auto const &r: rangemat[di].segments)
             for(int j = r.first; j < r.second; j++, i++)
-                tod[di*nsamp+j] = vals[i];
+                tod[(int64_t)(di*nsamp+j)] = vals[i];
```
mhasself: same.

src/array_ops.cxx
```diff
@@ -318,7 +318,7 @@ void pcut_poly_tod2vals_helper(const vector<RangesInt32> & rangemat, int resolut
                 for(int p = 0; p < np; p++) vals[i+p] = 0;
                 for(int s = r.first; s < r.second; s++) {
                     T x = -1 + 2*(s-r.first)/T(r.second-r.first-1);
-                    T t = tod[di*nsamp+s];
+                    T t = tod[(int64_t)(di*nsamp+s)];
```
mhasself: same

In all of these disjoint loops over detectors, I think it's simpler to compute a local pointer and use that.  The compiler will optimize it to the same thing, but it keeps things cleaner. i.e.
```
for(int64_t di = 0; di < rangemat.size(); di++) {
    auto this_signal = tod + di*nsamp;
    ...
```

src/array_ops.cxx
mhasself, line 359: missed this one?

mhasself, line 372: missed this one?

mhasself: replace with auto or int64_t.

mhasself, line 710: I'm not sure this is good enough -- need to cast di or nsamp explicitly.

src/array_ops.cxx
```diff
@@ -715,7 +715,7 @@ void _clean_flag(int* flag_data, int width, int ndet, int nsamp)
             if(det_flag[si]==0) {
                 // If this block was too small
                 if(count<width) {
-                    for(int i = si - count; i < si; i++){
+                    for(int64_t i = si - count; i < si; i++){
```
mhasself: Odd to change this type, but not `si`...

src/array_ops.cxx
mhasself, line 753: Cast reqd

mhasself, line 786: ding

mhasself, line 840: ding

mhasself, line 869: ding

mhasself, line 921: ding

mhasself, line 923: ding

mhasself: ding

src/array_ops.cxx
```diff
@@ -1133,7 +1133,7 @@ void _detrend(T* data, const int ndets, const int nsamps, const int row_stride,
     if (method == "mean") {
         #pragma omp parallel for num_threads(nthreads)
         for (int i = 0; i < ndets; ++i) {
-            int ioff = i * row_stride;
+            int64_t ioff = i * row_stride;
```
mhasself, line 1136: ding

src/array_ops.cxx
```diff
@@ -1153,7 +1153,7 @@ void _detrend(T* data, const int ndets, const int nsamps, const int row_stride,
     else if (method == "median") {
         #pragma omp parallel for num_threads(nthreads)
         for (int i = 0; i < ndets; ++i) {
-            int ioff = i * row_stride;
+            int64_t ioff = i * row_stride;
```
mhasself, line 1156: ding

src/array_ops.cxx
```diff
@@ -1185,7 +1185,7 @@ void _detrend(T* data, const int ndets, const int nsamps, const int row_stride,
 
         #pragma omp parallel for num_threads(nthreads)
         for (int i = 0; i < ndets; ++i) {
-            int ioff = i * row_stride;
+            int64_t ioff = i * row_stride;
```
mhasself, line 1188: ding

src/array_ops.cxx
mhasself, line 1232: ding

test/test_overflow_block.py
```diff
@@ -0,0 +1,15 @@
...
+import so3g
+import numpy as np
+
+
+class TestOverflowBlock(unittest.TestCase):
+
+    def test_overflow(self):
+        a = np.zeros((4050, 603260), dtype='float32')
+        so3g.block_minmax(a, a, 800, 2, 0)
+
```
mhasself, line 12: Can we add a bunch of unit tests for the "large array limit" of all the array_ops functions?  They could be disabled by default but it would be nice to run them, against this PR for example, or anytime we add new stuff.

mhasself (changes requested):
Since I started this, you said you knew it was incomplete.  But anyway, here are all the things I could find.

It does look like a lot of these inline casts would be unnecessary if we just did detector index loops as `for (int64_t di=0; ...`.  I'm totally ok with that approach.

Generally this is a huge pervasive problem, and it almost seems like the solution is "don't use `int` anymore, ever, for anything".  If I were overhauling this myself I would probably define a type to use for any integers that might get combined into array indices or offsets.  C++ has `ptrdiff_t` -- that would do the trick.

But also -- the parts of the code that are immune to this issue are instructive: if you make use of the BufferWrapper strides and pointer computation, these offset issues don't occur.  And as a bonus, you can handle slightly non-trivial data layouts more easily.

It's annoying to need to leave `int` behind! So easy to type ...

include/numpy_assist.h
skhrg: If we change `shape` below to be `int_64` then than is fine. I had avoided doing that because we had wanted to proceed with the assumption that the shape of the array should be limited s.t it is always `int` in each dimension. Hence `size` stays an `int` (and thus `ndet` and `nsamp` stay `int`) but `vshape` is allowed to be `int64_t` so that we can check it even if it is more than an `int`.

include/numpy_assist.h
skhrg: See above reply

